PR #17, opened 2018-10-02
Switch to icon svg images

bmansheim:


bmansheim:
Using png temporarily

PR #21, opened 2018-10-05
Fixed indentation issues on spec definitions

amiramm:


bmansheim:
@amiramm Thanks for your contribution!

PR #36, opened 2018-10-15
Suggest changing "eject" to "evict"

gavrie:


ushachar:
@bmansheim  seems like a good suggestion to me -- eviction is used in Redis as a term for deleting volatile keys,  but I think the distinction here is clear enough...

bmansheim:
@gavrie I'm going to leave this for now to not break the rule of not giving more than one meaning to a term.

gavrie:
> @bmansheim seems like a good suggestion to me -- eviction is used in Redis as a term for deleting volatile keys, but I think the distinction here is clear enough...

@ushachar OK, that explains why "ejection" was chosen since eviction was taken... I tend to agree with @bmansheim that overloading the term is not a good idea since we already have too much of that.

PR #83, opened 2018-11-13
Update using-oss-cluster-api.md

Paz-Yanover:
I'm not sure it's still required. Need to validate with Uri / Roey


ushachar:
@oridistor  - Can you check if this is still needed?

oridistor:
@ushachar this is not needed, the update Statemachine runs dmc reconf as part of its code.

PR #101, opened 2018-11-21
5.4 Add version selector to TOC

bmansheim:


bmansheim:
@HarunD To match the design we should also do this for all products:
1. Remove the arrow next to the top-level topic (instead of Home)
2. Add the line below the top-level topic
3. Add Expand all/Collapse all

image

PR #103, opened 2018-11-26
Update port-configurations.md

kerneljake:
TCP 53 is not needed for our DNS implementation.


bmansheim:
Thanks for the update @kerneljake ! I'll merge it after @Paz-Yanover gives the OK.

ushachar:
@bmansheim  - this is ok _for now_ 
There's a good chance we'll be using TCP based DNS in PCF environments soon....

PR #105, opened 2018-11-26
fix prometheus.yml indenting

jsachs-redislabs:
indentation is important for prometheus.yml. without the indentation, the prometheus.yml does not work.


bmansheim:
Yeah, I saw that before it was published but somehow the fix didn't get in.
Thanks @jsachs-redislabs  for fixing it up!

PR #114, opened 2018-12-04
Update quick-setup.md

ofirluzon:
Remove docker (non) related instructions


bmansheim:
@ofirluzon Good suggestion Ofir. I also merged the Docker instructions for each OS into one article.

PR #115, opened 2018-12-05
5.4 Add non-cluster CRDB changes

bmansheim:


bmansheim:
@Paz-Yanover Please review again. I reverted back to "You cannot enable or disable database clustering after the CRDB is created." because I couldn't find another clear way to say it.

PR #123, opened 2018-12-11
Add warning to avoid overwriting custom certificates

bmansheim:


bmansheim:
@bmansheim Add RED-23194 to commit message

PR #129, opened 2018-12-13
Version selector 5.2

bmansheim:


bmansheim:
Replace with updated PR

PR #134, opened 2018-12-19
Remove website search

bmansheim:


bmansheim:
@HarunD I removed the Search option in the menu but now clicking on the Login option opens the Search bar. Looks like some CSS needs to be removed.

PR #140, opened 2018-12-22
Add information to create pkg from command-line

vikrammoule-redislabs:
Add information to create pkg from command-line and also the debuginfo command is the entire package creation fails for some reason.


bmansheim:
@vikrammoule-redislabs Does this look correct?

vikrammoule-redislabs:
@bmansheim Yep. Looks fine.

PR #147, opened 2018-12-27
Add context text for search results

HarunD:
This adds a few words of context (3 before and 3 after the search term) for items displayed in the search results.


bmansheim:
@HarunD Please check a few search terms and see if the results make sense:
ssl
active-active
architecture
openshift
failover

Some of the things I saw were:

1. punctuation breaks context text
2. context text from summary pages only shows the searched word (if that)

PR #150, opened 2018-12-27
Force browser to use new version of style.css

bmansheim:


HarunD:
The changes introduced in this PR aim to make browsers load the newest stylesheet file (style.css) each time a user visits the website.
This is done by following steps:
1. Making the name of style.css unique by appending the hash of its contents, so the import will look like this:
`<link href="./theme-flex/style.6f300b8c332b51f8466de0da2d34d82cce44ada2c897ff810335e503e3a724d0.css" rel="stylesheet" integrity="sha256-bzALjDMrUfhGbeDaLTTYLM5EraLIl/&#43;BAzXlA&#43;OnJNA="> `
`integrity` param is added for added security and is used to confirm that the data has not been tampered with.
2. In order to obtain the hash of the file, a built-in method in Hugo is used called `resources.Fingerprint`
This method is used in `head.html` partial in the following way:
` {{ $style := resources.Get "theme-flex/style.css"  | resources.Fingerprint }}
   <link href="{{$style.RelPermalink}}" rel="stylesheet" integrity="{{$style.Data.Integrity}}">`
3. Once style.css is edited, two more files will be generated for the hash and integrity mechanism to work (These files need to be committed and pushed along with style.css.):
screenshot 2018-12-27 at 12 48 02

4. A `.htaccess` file is added in the root of the project with the goal of deliberately avoiding any cacheing of `style.css` file as agreed with Ben. This will be the case in the transition period of couple of weeks and can be reverted.

PR #159, opened 2019-01-07
Add "Expand All / Collapse All" toggle in the menu

HarunD:
This change adds a toggle for showing and hiding of menu items with single click.
The default state when the page is loaded to `/rs`, `/rv` or `/rc`, ie. the root page of each section, is that the menu is collapsed and the toggle text is "Expand All" with the appropriate icon.

The change should be visible [here](https://docs.redislabs.com/staging/expand-collapse) but for some reason it is returning 404 even though builds in CircleCI have been successful.

### DIFF
--- a/layouts/partials/flex/body-beforecontent.html
+++ b/layouts/partials/flex/body-beforecontent.html
@@ -1,7 +1,12 @@
 
+{{ if eq (getenv "HUGO_ENV") "production" }}
+{{ with .Site.Params.googleTagManager }}
+  {{ partial "gtm-body" . | safeHTML }}
+{{ end }}
+{{ end }}
 <header>
-  <div class="logo">
-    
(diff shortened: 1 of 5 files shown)


HarunD:
@bmansheim 
1. I aligned the version dropdown to be in the same level as Expand/Collapse
2. I managed to change the toggle to behave like that, to control the visibility of first-level child items. But through testing I found some cases where it does not behave as needed due to the multi-level structure of the menu. This can be done but I see now that it will take a bit more time.

bmansheim:
@HarunD I see that you did some work on this. The current state is incomplete. What I see is:
expand-collapse

bmansheim:
Replaced by https://github.com/RedisLabs/redislabs-docs/pull/363

PR #161, opened 2019-01-08
Change footer year

bmansheim:


bmansheim:
@HarunD Nice to see that I did the right thing just in the wrong place. 😄

PR #164, opened 2019-01-13
RED-24561 Configuration instructions for TLS authentication 5.4.2

bmansheim:


Paz-Yanover:
For 5.4.2- please update the SSL mentions to use TLS or SSL/TLS and probably also the screenshots.

PR #165, opened 2019-01-13
RED-20882 Add missing metrics definitions

bmansheim:


bmansheim:
@Paz-Yanover @ushachar Are we ready to go with these updates?

PR #166, opened 2019-01-14
Add note about upgrade mode (and style fixes)

bmansheim:


content/rs/installing-upgrading/upgrading.md
```diff
@@ -40,13 +40,16 @@ You will run install.sh from the directory where you untarred the media
...
-Just like for a new installation, you must sudo to or be root to do the
+Just like for a new installation, you must sudo or be root to do the
 upgrade.
 
 ```src
 $ sudo ./install.sh
 ```
 
+**Note**: Make sure that you answer Y to questions about upgrade mode so that the
+upgrade mode is enabled during the upgrade process.
```
bmansheim: @Paz-Yanover Please approve the note or add comments.

Paz-Yanover: I think we'd better push the specific questions to be more clear.
In addition, I think we need to be more vague with the explanation- because it sounds like we are stupid...

bmansheim: @Paz-Yanover How's this?
Make sure that you answer Y to these questions:
- At the beginning of the upgrade process - "Do you want to enable upgrade mode now [Y/N]?"
- At the end of the upgrade process - "Do you want to disable upgrade mode now [Y/N]?"

PR #172, opened 2019-01-16
Improve explanation of moving socket files

bjbredis:


bmansheim:
@ushachar Is this page still relevant? If yes, do the socket files get moved out of /tmp during upgrade?

ushachar:
---> @RoeyPrat

RoeyPrat:
> @ushachar Is this page still relevant? If yes, do the socket files get moved out of /tmp during upgrade?

1. the page is still relevant. 
2. before RS version 4.5.0-51, socket could not be moved.
3. in RS version 4.5.0-51 and above, socket can be moved manually, with the process described here.
4. starting with version 5.2.2, in **new** installations of cluster, the socket path defaults to /var/opt/redislabs/run.
5. socket path is not changed automatically in upgrade.
6. if you upgrade from before version 5.2.2 to it or a later version, socket path stays what is was.
Specifically, if socket path was /tmp, it stays /tmp unless changed manually.
=================
@bjbredis great PR! I would prefer a paragraph about the change of 4.5.0-51, and a separate paragraph about the change of 5.2.2. I think writing about the `-s` option and these two changes in one paragraph is confusing.

thanks!

bmansheim:
@RoeyPrat Please review my changes. I don't mention that it is not possible to change the socket location in 4.5.0-51 or below because we are assuming that the customer is currently running 5.4.

content/rs/installing-upgrading/configuring/change-location-socket-files.md
```diff
@@ -6,66 +6,62 @@ alwaysopen: false
...
-## Changing the Location on an Existing Redis Enterprise Software (RS) Installation
+- `/tmp` - In clean installations of RS version lower than 5.2.2
+- `/var/opt/redislabs/run` - In clean installations of RS version 5.2.2 and higher
 
-You must be running RS version 4.5.0-51 or higher, and with the sock
-files location is still under the default location of /tmp. (This is
-where they go when install.sh is executed without -s flag.)
+When you upgrade from a RS version lower than 5.2.2 to 5.2.2 and higher, the socket files
+are not moved to the new location by default. During [installation]({{< relref "/rs/installing-upgrading/downloading-installing.md" >}})
+you can specify a custom location for the socket files, but after installation
```
Paz-Yanover: @RoeyPrat - I think we need to add the reason and the motivation for changing the socket path. Do you know what's the reason for it?

content/rs/installing-upgrading/configuring/change-location-socket-files.md
```diff
@@ -6,66 +6,62 @@ alwaysopen: false
...
+        {{% /warning %}}
         ```src
         $ rladmin restart db <db name>
         ```
 
-    1. Execute the following three commands **for each** master/slave
-        shards during a maintenance window. All three of these commands
-        **must** be executed **on the node hosting the shard being
-        operated on**:
+    - For each master/slave shard for each database in the cluster, run:
```
Paz-Yanover: @RoeyPrat  - what's the difference between this way and the previous one? do you know why we have both? It seems to me that the first one should be enough

content/rs/installing-upgrading/configuring/change-location-socket-files.md
Paz-Yanover: --> Some customers have procedures which delete the /tmp content. So 
the motivation is some customers' operations which deletes all files under /tmp
@bmansheim

content/rs/installing-upgrading/configuring/change-location-socket-files.md
Paz-Yanover: @bmansheim - also @RoeyPrat doesn't see any reason for having the second option- option #1 should be good enough. However, it worth checking who performed the original documentation commit and get more input.
BTW- @RoeyPrat  suggested to write the shards restart procedure/s in a separate page and add reference from all the relevant pages (like this one). I think it can be useful in case we have more than one place that explains this procedure. Otherwise, I'm not sure it worth the efforts (for now)

content/rs/installing-upgrading/configuring/change-location-socket-files.md
bmansheim: @Paz-Yanover Added

content/rs/installing-upgrading/configuring/change-location-socket-files.md
bmansheim: @Paz-Yanover This content was brought from WordPress so I don't know who added it. Do you want to send it to someone else to review, like QA?

content/rs/installing-upgrading/configuring/change-location-socket-files.md
Paz-Yanover: @bmansheim - good, thanks

content/rs/installing-upgrading/configuring/change-location-socket-files.md
Paz-Yanover: I don't think QA is relevant here. Maybe @ushachar  has some insights here.

bmansheim: @ushachar Should we include the "Restart each master and slave shard" instructions or just give the "Restart each database" instructions?

ushachar: I think we should stay with 'each master & slave' - otherwise folks might try and use the rladmin restart....

bmansheim: @Paz-Yanover I clarified with Uri. He said Roey is the authoritative answer. I'll merge the PR.

Paz-Yanover: @ushachar - I didn't understand your comment. The original documentation talks about two options to "restart all of the shards in the cluster": (1) by using "rladmin restart db" (2) by using redis_ctl stop; rladmin failover shard; redis_ctl stop
Do you think that we need also option #2 and the first option is not enough / good?

ushachar: As I told @bmansheim today - @RoeyPrat has the authoritative answer on this as he implemented the feature and should know what's needed...

Paz-Yanover: @RoeyPrat @bmansheim - so, can we remove the second suggestion to restart all the shards in the cluster?

bmansheim: @Paz-Yanover That's my understanding. I removed the instructions for restarting all shards.

RoeyPrat: yes. 😆

PR #178, opened 2019-01-23
Fix landing page links

bmansheim:


bmansheim:
@HarunD .../latest/rc goes to 404 but ...latest/rs works. Any idea why?

If not then this PR is to change all links on the landing page to have a trailing slash.

HarunD:
@bmansheim I tested the [landing page](https://docs.redislabs.com/latest/index.html) in several browsers and mobile Chrome as well, the link on landing page to `/latest/rc` works.
Are you still able to reproduce the 404?

bmansheim:
@HarunD Still /latest/rc still gives me a 404 but /staging/landing-links/rc doesn't. Don't know.

PR #182, opened 2019-01-28
RED-21842 Improve warning of upgrade order

bmansheim:


bmansheim:
@laurentdroin I reverted your commit and suggested this one instead in consultation with @Paz-Yanover .

laurentdroin:
@bmansheim - I am not sure the new text is correct. I think that after upgrading the master node, if the master node goes down (causing a new master to be elected) and then back up, you may not be able to upgrade any node (not even the new master). This is certainly the case when upgrading to 5.2.0 or above from below 5.0.2 or below. I think we need to insist that no node goes down while an upgrade is ongoing. Maybe we check this with @ushachar.

ushachar:
@laurentdroin  - As much as I'd wish for it, this isn't something we can insist on in any way - nodes will go down during upgrade, and the cluster has to be able to deal with it.

laurentdroin:
@ushachar OK but still, the newly proposed text doesn't work because the newly elected master node may still not be upgraded (if tasks in the new version are submitted to the new master which is in the old version).
Our scenario is this:
Upgrade the master from 5.0.2 to 5.2.0 => fine
Shutdown this master => a new master is elected, which is still in 5.0.2
The old master (now a slave) that is at 5.2.0 is restarted. In 5.2.0, resource_mgr, at start time, submit a RetryBDB task to the new master who is in 5.0.2 and who doesn't know how to process it. The task gets stuck.
Trying to upgrade the new master will fail because the upgrade will stop due ti background processes.

What needs to be done is either to remove the RetryBDB task from the system (this is what we have been doing and it's fairly safe)
or maybe - I have not tried this for fear of causing more issues - set the already upgraded node as the master and let it process the task.

But I am all for letting the cluster handle the situation by itself. This  will require some code changes and it is too late for all the customers upgrading from 5.02 and below to 5.2.0 and above.
I think modifying the doc (maybe until 5.2.0 is obsolete) to tell customers to not restart a node doing an ongoing upgrade would be a good solution for a while. We have ad at least 3 or 4 support tickets about this.

bmansheim:
@Paz-Yanover What do you think?

Paz-Yanover:
@bmansheim @laurentdroin - we can add a recommendation to avoid 'manual' node restart during the upgrade process in order to speed up the upgrade process or such. But- we can't say that node restart may cause problems, it looks bad

bmansheim:
@Paz-Yanover Something like: "We recommend that you plan to keep all nodes up until the upgrade is completed on all nodes."

Do you want this in place of the note about the master node going down suggested in this PR?

bmansheim:
@laurentdroin I updated the note according to the comment from Paz. Looks OK to you?

laurentdroin:
@bmansheim Yes. I think this is good. Thank you.

PR #188, opened 2019-01-31
Add next steps to RS install instructions

bmansheim:


content/rs/installing-upgrading/downloading-installing.md
```diff
@@ -261,3 +261,11 @@ It is advisable to answer 'Y' on this. If
 you would like to see the contents of this script, it is installed
 to /opt/redislabs/bin/rlcheck on every Redis Enterprise
 node.
+
+## Next Steps
+
+After you install RS, you can:
+
+1. [Create a cluster]({{< relref "/rs/administering/cluster-operations/new-cluster-setup.md" >}})
```
Paz-Yanover: Actually, as written in step 6 above, after you install RS, you 'must' *create or join* a cluster. That's the next step.
Then, you can create a database / CRDB (- good addition to the documentation).

bmansheim: Changed it according to your comment

PR #192, opened 2019-02-05
RED-21717 VPC peering

bmansheim:


bmansheim:
Also covers https://redislabs.atlassian.net/browse/RED-23035

PR #197, opened 2019-02-07
multi proxy is confusing the customers

GitaShmueli:
The section about multi proxy is confusing the customers, as they think to move to multi proxy configuration just for the upgrade purpose.

### DIFF
--- a/content/rs/installing-upgrading/upgrading.md
+++ b/content/rs/installing-upgrading/upgrading.md
@@ -56,9 +56,10 @@ $ sudo ./install.sh
 
 {{% /note %}}
 
-During the node upgrade process, the services running RS are restarted.
-This will result in a very short interruption to the proxy and node
-itself and the databases running on it.
+The node upgrade process restarts the services running RS, which causes
+a short interruption to connections to the proxy, node and databases.
+Clusters with [multiple active proxies]({{< relref "/rs/administering/designing-production/networking/multiple-active-proxy.md" >}})
+have reduced interruption time.
 
 **Note**: In order to ensure cluster and databases' availability, it is
 important to upgrade the nodes one by one, and **not attempt to upgrade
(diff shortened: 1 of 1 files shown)


content/rs/installing-upgrading/upgrading.md
```diff
@@ -58,11 +58,7 @@ $ sudo ./install.sh
 
 During the node upgrade process, the services running RS are restarted.
 This will result in a very short interruption to the proxy and node
-itself and the databases running on it. This can be dramatically reduced
-by using multiple active proxies for your database. The client will lose
-the connection to the proxy on the node being upgraded, but then
-automatically reconnect to another active proxy you have configured for
-the database.
+itself and the databases running on it.
```
Paz-Yanover: I think we need to mention here that in case of working with multiple proxies, the interruption will be minimal. 
Gita didn't like the fact that it sounds like we recommend to move to multiple proxies before the upgrade (which makes no sense at all).
Let's try to rephrase it and get Gita's approval

content/rs/installing-upgrading/upgrading.md
```diff
@@ -56,13 +56,10 @@ $ sudo ./install.sh
...
-This will result in a very short interruption to the proxy and node
-itself and the databases running on it. This can be dramatically reduced
-by using multiple active proxies for your database. The client will lose
-the connection to the proxy on the node being upgraded, but then
-automatically reconnect to another active proxy you have configured for
-the database.
+The node upgrade process restarts the services running RS, which causes
+a short interruption to connections to the proxy, node and databases.
+Clusters with [multiple active proxies]({{< relref "/rs/administering/designing-production/networking/multiple-active-proxy.md" >}})
+have reduced interruption time.
```
bmansheim: @Paz-Yanover @GitaShmueli What do you think of this?

Paz-Yanover: After discussing it f2f with @GitaShmueli  and Miki, and based on their Support experience, let's accept Gita's PR and remove the multi-proxy mention at all.

PR #203, opened 2019-02-11
Fix page title anchor position so it is not covered by the header (#200)

bmansheim:
@HarunD Please look at the behavior of links in the page contents


bmansheim:
https://github.com/RedisLabs/redislabs-docs/pull/215

PR #205, opened 2019-02-11
Add style for images that should not be clickable

HarunD:


bmansheim:
@HarunD How do we apply this style to the icons?
You can use this page to see how they look: https://docs.redislabs.com/staging/icon-images/rv/administration/setup_and_editing/view-edit-subscription/

HarunD:
@bmansheim sorry, I forgot to add it in the PR description.
To apply the style you need to define the icon like this:
`icon_add`

Note the **#no-click** at the end of the icon path. With this we can target specific links in CSS and style them.

bmansheim:
@HarunD I tried but it didn't work for me. Try find *icon_(.*)png* and replace with *icon_$1png#no-click*.

HarunD:
@bmansheim I will take a look at that. Thanks for letting me know.

PR #206, opened 2019-02-12
Add ldap_bind_dn to example saslauthd.conf file

kerneljake:
make the two example config files match


bmansheim:
@danni-m Can you approve this?

danni-m:
@bmansheim yes, but I don't know enough to say if that's the correct way. I guess that @kerneljake verified it?

PR #211, opened 2019-02-13
Separate platforms section

bmansheim:


bmansheim:
@HarunD Here are a few fixes:
1. The search dropdown should be the same size as the search box.
2. When I click on the icons I see the icon file. It should go to the Overview page of that section.
3. The search dropdown should show about 5 search results. No need to have it drop past the bottom of the page.
image

bmansheim:
@HarunD Also, we might need to tweak the sizes of the icons so that they are all the same size as the Cloud Pro icon. How much work would that be?

image

bmansheim:
Approved by @Paz-Yanover

PR #216, opened 2019-02-14
RED-25830 Specify new CRDB with same version

bmansheim:


bmansheim:
@Paz-Yanover I didn't see any articles about adding a node to CRDB

PR #220, opened 2019-02-19
Update rs-5-4-december-2018.md

GitaShmueli:


bmansheim:
@GitaShmueli Good idea. I moved the reference to be a link of the paragraph text. And, you almost got the relref syntax correct!

PR #221, opened 2019-02-19
Sync HTML and CSS files with max-age=0

bmansheim:


bmansheim:
@HarunD You can add the autogen stylesheet in this branch.
It is published to https://docs.redislabs.com/staging2/set-html-cache-0. The html and css files should not be cached so changing the stylesheet should be reflected immediately.

bmansheim:
This PR was for testing. Change implemented in production in: https://github.com/RedisLabs/redislabs-docs/pull/295

PR #222, opened 2019-02-19
Rewrite Add Nodes

bmansheim:


bmansheim:
@Paz-Yanover When working on RED-25830 I rewrote this whole topic before I realized it's not CRDB.

Let me know if you have any reservations about publishing it.
https://docs.redislabs.com/staging/rewrite-add-nodes/rs/administering/cluster-operations/adding-node/

PR #227, opened 2019-02-21
Add port 9443 for node bootstrap

zoharfuchs:
Port 9443 is used for bootstarp. See JIRA 26312


bmansheim:
@ushachar I think it depends on if people are aware that node bootstrapping works via the API. If yes, then no need to mention. If no and people will need to know which port it works on then we can add:
| 8080, 9443 | Used to expose the REST API, including cluster management and node bootstrap |

ushachar:
@bmansheim - I don't know what's 'obvious enough' here -- so let's go with your phrasing...

bmansheim:
Moved into https://github.com/RedisLabs/redislabs-docs/pull/233

PR #231, opened 2019-02-27
Add table of RS hot topics

bmansheim:


bmansheim:
Currently we have a few pages that get a lot of traffic but have little function. You can see from the attached analysis that a large portion of our hits are on the RS main page but that page has very little useful content.
[Analytics docs.redislabs.com Pages 20190218-20190319.pdf](https://github.com/RedisLabs/redislabs-docs/files/2988194/Analytics.docs.redislabs.com.Pages.20190218-20190319.pdf)
This change is to add a table of top pages so that users that get to RS main page can easily find the most popular pages on the site.
@Paz-Yanover Does the table look good enough or does it need design help before we push it live?

bmansheim:
@HarunD Looking good!
Please check these details:

- Enlarge heading topics
- Reduce space between sub-topics
- Extend Hot Topics box across page and push Page Contents down (in other words, the box is connected to the heading 1 as opposed to the article body)
- Enforce spacing between columns so that the links break also in wider screens

Documentation_ArticlePage_1280_HotTopics

bmansheim:
@HarunD Looks great now.
Can you add space on the right at 1280 pixels width so that the margin on each side is the same like it is in the design?

bmansheim:
Also, how can do I reproduce it if I want to do the same thing on another page?

HarunD:
> Also, how can do I reproduce it if I want to do the same thing on another page?

In order to display the table on other pages, this piece of code in `body-beforecontent.html` file should be modified:
```
        {{if eq .RelPermalink "/rs/"}}
          {{ partial "rs-hot-topics-table.html" . }}
        {{end}} 
```

So, for example, if we want to show the table on top of each article within RS/Software, this code would look like this:
```
        {{if .Params.categories}}
          {{if eq (index .Params.categories 0) "RS"}}
            {{ partial "rs-hot-topics-table.html" . }}
          {{end}}
        {{end}} 
```

bmansheim:
Perfect!!

bmansheim:
@HarunD I implemented a few changes for line spacing and margins. I also saw that in mobile the right arrow is again covering the title. I think that will be the last change.
image

bmansheim:
@HarunD Ignore that. I see that it was resolved in another PR.

bmansheim:
@HarunD I rebased the branch and it looks good but I still see the right arrow overlapping with the title in mobile screen size (360).

bmansheim:
@HarunD That was easy, eh?

PR #235, opened 2019-03-05
Embed Shortcode

HarunD:
This PR moves RBAC table from Account Management to a partial file and import it back via **readFile** shortcode.


bmansheim:
@HarunD That's really good to know. I forgot to mention that the table is the HTML table in the RED-24044-update-db-viewer branch. How would you do the same with the HTML table?
Also would a shortcode make it a simpler syntax? For example {{% embed-md "rs/partials/rbac-table.md" %}}

HarunD:
@bmansheim I will add the same for HTML table you mentioned, and will also create two shortcodes for simpler syntax (one for md and one for HTML content).

HarunD:
@bmansheim I added shortcodes for md and HTML, with simple usage syntax. They are used like this:
```
{{< embed-md "rs/partials/rbac-table.md" >}}
{{< embed-html "rs/partials/sample-table.html" >}}
```

[Here](https://docs.redislabs.com/staging/embed-shortcode/rs/administering/security/account-management/) you can see how this table looks.

If this looks good I will remove the testing file **sample-table.html** and its reference in **account-management.md.**

bmansheim:
@HarunD Looks great. One comment: the relref and images syntax both include the leading / so maybe it would be good to keep that syntax consistent, if possible.

bmansheim:
@HarunD I knew I shouldn't have said "one comment". 😄 
Probably best to keep these partials outside of the /content path so:

1. they don't get published into the site
2. so they are not associated with specific products (rs, rc, rv, etc)

HarunD:
@bmansheim I've added the forward slash so the syntax is now consistent and have also moved the partials out of the `content` folder.
The `sample-table.html` is still there for testing, so once everything looks good to you let me know and I will delete that file.

bmansheim:
@HarunD I think you can remove the sample-table.

HarunD:
@bmansheim sample-table removed

PR #274, opened 2019-03-11
Update create-subscription.md

allegrass:


allegrass:
Merged with prior changes

PR #289, opened 2019-03-12
the original link is not working

GitaShmueli:


bmansheim:
I'll look for other instances and try to fix them also.

PR #292, opened 2019-03-13
RED-22752 Sync shards in parallel

bmansheim:


bmansheim:
We agreed that we will add an explanation of gradual_sync when we release the rladmin docs.

PR #294, opened 2019-03-14
Map port 9443 in Docker getting started

laurentdroin:
Need to also map port 9443.


Paz-Yanover:
@ushachar - is it a must to add port 9443 while running RE on docker?

ushachar:
Not an absolute must, but much more convenient.

bmansheim:
@ushachar If this is the only other port that someone might want then I'll add it. If someone might want to add other ports also, I suggest that instead we write: "For other ports to open, see "Network Configuration". What do you think? (cc: @Paz-Yanover )

bmansheim:
@ushachar @Paz-Yanover I'll add the port and a link to the list of ports in case customers want others.

PR #302, opened 2019-03-25
RED-26845 Update for register-counter changes in 5.4.2

bmansheim:


bmansheim:
@Paz-Yanover Waiting for details on upgrade process

content/rs/installing-upgrading/upgrading.md
```diff
@@ -120,3 +120,19 @@ a result:
     usually takes longer than an RDB file.
 - For databases that have neither replication nor [persistence]({{< relref "/rs/concepts/data-access/persistence.md" >}})
     enabled, the database loses all its data after it is restarted.
+
+## Upgrading CRDB
+
+When you upgrade from RS 5.4.0 or lower to 5.4.2 or higher, the upgrade includes fundamental changes in the CRDB protocol so you must upgrade your CRDBs to use the new CRDB capabilities.
+
+After you [upgrade your existing nodes](#upgrading-nodes) in your cluster to RS 5.4.2, the status of your existing CRDB is displayed with ‘OLD CRDB PROTOCOL VERSION’. The shard status shows that an ‘OLD REDIS VERSION’ is used. After you upgrade the CRDB, the new Redis version is used and the status is updated accordingly.
```
Paz-Yanover: We must say the following:
CRDB protocol is backward-compatible, which means that RS 5.4.2 CRDB instances that use the new protocol can understand write-operations that come from instances on RS versions below 5.4.2. However, the protocol is not forward-compatible, so CRDB instances with the old protocol cannot understand write-operations of instances with the newer protocol version.
As a result, after you upgrade the CRDB protocol on one instance, other instances that were not yet upgraded may not be able to sync from the upgraded instance. The upgraded instance receives updates from upgraded and non-upgraded instances.
Therefore, we highly recommend that you upgrade all instances of a specific CRDB within a reasonable time frame and avoid temporary inconsistencies between the instances.

Paz-Yanover:
@guybe7 @vadim@redislabs.com @yuval@redislabs.com - can you please review https://github.com/RedisLabs/redislabs-docs/pull/302/files#diff-cbe49f31c479c315e3f47b59078f2fed (all the file) to make sure it's correct after the hybrid counter-register change?

bmansheim:
@Paz-Yanover No responses yet. Do we wait or merge?

Paz-Yanover:
Let's wait. I asked Yuval to review it (Slack). We need him / @guybe7 to review it

PR #309, opened 2019-04-01
Document working shortcodes

HarunD:
This change adds documentation for working shortcodes to the cheatsheet file.
Working shortcodes have also been copied from docdock's shortcode directory to `/layouts/shortcodes`.

I have added descriptions for new shortcodes in the file but I suggest @bmansheim you review those as well and make any style changes needed.


content/cheatsheet.md
```diff
@@ -81,6 +81,20 @@ def hello_world():
   print "Hello World!"
 ```
 
+## Code
+
+Code can be displayed using the code shortcode as well.
```
bmansheim: Is there a functional difference between this and code fences?

bmansheim:
I think we can get rid of these because they don't seem so useful:
- Alert
- Anchor Link
- Button
- Icon
- Emoticon
- Header
- TOC

content/cheatsheet.md
```diff
@@ -116,6 +130,57 @@ To make an image appear on the next line in a list:
 1. Add 2 spaces at the end of the line and put the image on the next line with a tab indentation.  
     ![Redis Enterrpise Cluster]( /images/rs/rp_stack.png )
 
+### Image shortcode
```
bmansheim: Is there a functional difference between this and  ?

bmansheim:
The excerpts don't render correctly now. https://docs.redislabs.com/staging/shortcodes-cleanup/cheatsheet/#excerpts

HarunD:
> I think we can get rid of these because they don't seem so useful:
> 
> * Alert
> * Anchor Link
> * Button
> * Icon
> * Emoticon
> * Header
> * TOC

I agree. I will remove those from the cheatsheet.

content/cheatsheet.md
HarunD: I believe there is no difference. Personally, code fences are more readable so I suggest those are used (as they are now).

I documented the **code** shortcode so we have all the working shortcodes in one place and can remove it if you agree.

content/cheatsheet.md
HarunD: Image shortcode (ie. {{% image..}}) can only display images from the site itself, ie. images located in _static_ directory (or its subdirectories), whereas the other way (ie. Redis Enterrpise Cluster) can display both those images and external images from other domains.

Therefore I think that the Redis Enterrpise Cluster option is sufficient as it covers both local and external images.

(similar to code shortcode, this was documented to have all working shortcodes in one place and can be removed from the cheatsheet).

HarunD:
> The excerpts don't render correctly now. https://docs.redislabs.com/staging/shortcodes-cleanup/cheatsheet/#excerpts

I see that the **image** shortcode actually messes up the rendering of excerpt shortcodes. When I remove it from the cheatsheet excerpts render correctly. Therefore I suggest we remove **image** shortcode as it is not really needed (as described in above comments).

content/cheatsheet.md
bmansheim: 👍

PR #311, opened 2019-04-02
RED-23377 Maintenance node refuses shard migration

bmansheim:


content/rs/administering/cluster-operations/maintenance-mode.md
```diff
@@ -17,6 +17,8 @@ When you turn maintenance mode on, RS:
 1. Takes a snapshot of the node configuration as a record of which shards and endpoints are on node at that time.
 1. Marks the node as a quorum node to prevent shards and endpoints from migrating into the node.
 1. Migrates shards to other nodes and binds endpoints to other nodes, if space is available on other nodes.
+1. Refuses requests for shard migration to the maintenance node.
```
RoeyPrat: this is very similar to the 3rd point ("Marks the node as a quorum node...").
perhaps add this comment to that point?

bmansheim: Good catch.

content/rs/administering/cluster-operations/maintenance-mode.md
```diff
@@ -16,6 +16,7 @@ When you turn maintenance mode on, RS:
 1. Checks whether shutdown of the node causes quorum loss. If so, maintenance mode is not turned on.
 1. Takes a snapshot of the node configuration as a record of which shards and endpoints are on node at that time.
 1. Marks the node as a quorum node to prevent shards and endpoints from migrating into the node.
+    The maintenance node entry in the rladmin status output is colored yellow to indicate that it cannot accept shard migration.
```
Paz-Yanover: (1) I think we'd better add a screenshot.
(2) We need to emphasise here (and maybe in another place where we're talking about the quorum node) that this applies to any node which cannot accept shard migration- i.e. maintenance mode, but also a quorum node.

bmansheim: OK.
We don't have discussion of quorum nodes in the docs. It really needs to be mentioned in a new section that talks about quorum nodes, not here.

Paz-Yanover: But we do need to say here that it's relevant for quorum node and not only node in a maintenance mode, don't you think?

PR #312, opened 2019-04-02
Theme cleanup

HarunD:
This change moves all the files edited directly in _themes/docdock_ out of this directory so it now represents the original docdock theme without any modifications.

All the modifications are now done in _static_ and _layouts_ directories in the root of the site.


bmansheim:
@HarunD Waiting for rebase (Not urgent)

PR #316, opened 2019-04-10
add "hostname" to URL format

hongliuliao:
The rdb file path format lost hostname , i add "hostname" to it


bmansheim:
Thanks for that contribution, @hongliuliao !

PR #319, opened 2019-04-12
Fix a minor typo

ghimire:


bmansheim:
@ghimire Thanks for helping with this small mistake. The little things count a lot.

PR #333, opened 2019-04-22
Import from another RedisLabs DB on Redis Essentials is not supported

vikrammoule-redislabs:
** Please note import from another RedisLabs DB on Redis Essentials is not supported.


bmansheim:
@allegrass @Aviad-Abutbul Please approve the change.

PR #343, opened 2019-04-30
Update persistence.md

AlonMagrafta:


bmansheim:
@AlonMagrafta Thanks for the fix!

PR #347, opened 2019-05-01
/images/rs/crdb-tls-all.png is not correct

GitaShmueli:
need new screenshot when "TLS for All communications" is chosen


bmansheim:
Resolved in https://github.com/RedisLabs/redislabs-docs/pull/383

PR #349, opened 2019-05-01
Add parallel_shards_upgrade and preserve_roles to upgrade procedure

OrenKarmi:
updating recent suggestion

### DIFF
--- a/content/rs/installing-upgrading/upgrading.md
+++ b/content/rs/installing-upgrading/upgrading.md
@@ -102,6 +102,11 @@ RS, Redis Labs recommends that you upgrade your Redis databases.
     You cannot upgrade databases before all of the nodes in the cluster are upgraded.
 1. In the [rladmin CLI]({{< relref "/rs/references/cli-reference/rladmin.md" >}})
     on any node in the cluster, run this command for each database: `rladmin upgrade db <db-name>`
+    This commands upgrades all DBs at the same time. When upgrading a multi DBs system, it is
+    advised to limit the number of concurrent DB upgrades with the parallel_shards_upgrade option.
+    It is also possible to preseve the original shard roles (master, slave) after upgrade, with
+    the preserve_roles option. `rladmin upgrade db <db-name> parallel_shards_upgrade 2 preserve_roles`
+    
 
 During the database upgrade process, the database is restarted. As
 a result:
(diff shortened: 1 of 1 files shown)


OrenKarmi:
done editing

bmansheim:
@Paz-Yanover Do we want to add this information to the upgrade instructions?

Paz-Yanover:
@bmansheim - yes, we can this information. We just need to verify that it's accuracy- e.g. whether by default all the shards are upgraded on the same time. Anyway, we should mention there that slave and master shards are not upgraded "together" (- as otherwise we would have data loss every upgrade...).
In addition, we need to think how to present these options- ux-wise

OrenKarmi:
@Paz-Yanover @bmansheim , this issue is requested (and resolved) here:
https://redislabs.atlassian.net/browse/RED-18875
(config "rladmin upgrade db" to do it gradually shard by shard)

bmansheim:
@RoeyPrat 

1. Can you confirm that this change is accurate? 
1. Obviously the shards aren't upgraded at the same time because master and slave must be upgraded separately, so how does it work?

RoeyPrat:
1. Not exactly. 
[Here](https://github.com/RedisLabs/redislabs-docs/pull/349/files#diff-cbe49f31c479c315e3f47b59078f2fedR105) we state that all DB shards at the same time.
1a. By default we restart all **slave** shards at the same time, as @Paz-Yanover explains [above](https://github.com/RedisLabs/redislabs-docs/pull/349#issuecomment-488606897).
1b. The **default** number of shards restarted simultaneously can be changed cluster-wide. The default it is "all", but we can edit this using `rladmin tune cluster parallel_shards_upgrade  <int>`. To view the current setting, run `rladmin info cluster | grep parallel_shards_upgrade`

2. Slaves are upgraded, X at a time, to completion. Then we fail-over, and then we upgrade the shards that have just been demoted. Then, and only if `preserve_roles` specified, we fail-over again, to return the demoted masters to their original role. 

@galcohen-redislabs you moved this logic out of rladmin during RED-14158. Please specify any changes in behavior if there were any.

galcohen-redislabs:
No change in behavior was done when moving the code away from rladmin, but I'll add a refinement to 1:
1c. If bdb isn't replicated, or if the force_discard flag was set - there will be no failovers at all.

Paz-Yanover:
@bmansheim - the information that @galcohen-redislabs added is important, but I don't think these options should be mentioned in our (external) documentation (- maybe only mention that failover is performed in case replication is defined or such).
Such information should appear in our (future) internal documentation...

bmansheim:
@Paz-Yanover There was already a description of the upgrade process for databases with replication so I improved it. Is that enough?

PR #368, opened 2019-05-13
Prevent "Hot topics" from covering search results

bmansheim:


bmansheim:
@HarunD When you enter a search parameter on the /rs page, the Hot Topics title overlays on top of the search results. I think this fixes the issue.

PR #376, opened 2019-05-15
Add key file for SFTP and backup rotation note for mount point

tom-andrews:
Added key file location for SFTP server configuration.
Added (lack of) backup rotation note for Local mount point.

### DIFF
--- a/layouts/partials/embeds/backup-locations.md
+++ b/layouts/partials/embeds/backup-locations.md
@@ -23,7 +23,7 @@ Before you specify to backup to an SFTP server, make sure that:
     - **Use the cluster auto generated key** - Go to **settings** and copy the **Cluster SSH Public Key**
         to the SFTP server.
     - **Use a custom key** - Generate a TLS key pair for the SFTP server, copy the private key to
-        the **SSH Private Key** box, and copy the public key to location required by the SFTP server.
+        the **SSH Private Key** box, and copy the public key to the location required by the SFTP server.
 
 To backup to an SFTP server, enter the SFTP server location in the format:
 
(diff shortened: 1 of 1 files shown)


layouts/partials/embeds/backup-locations.md
```diff
@@ -24,6 +24,7 @@ Before you specify to backup to an SFTP server, make sure that:
         to the SFTP server.
     - **Use a custom key** - Generate a TLS key pair for the SFTP server, copy the private key to
         the **SSH Private Key** box, and copy the public key to the SFTP server.
+    - The key shoud be added to ~/.ssh/authorized_keys of the user used for backup on SFTP server.
```
bmansheim: @ushachar Should we tell customers where to put the key?

layouts/partials/embeds/backup-locations.md
```diff
@@ -48,6 +49,7 @@ Before you specify to backup to a local mount point, make sure that:
 - The node has network connectivity to the destination server of the mount point.
 - The `redislabs:redislabs` user has read and write priviledges on the local mount point
 and on the destination server.
+- Periodic backup doesn't include rotation functionality. It's your responsibility to enforce and maintain backup retention policy.
```
bmansheim: @RoeyPrat Each backup file has a unique name, right?

layouts/partials/embeds/backup-locations.md
ushachar: No - it depends on which SFTP server they are running, and not something we can control / know.

layouts/partials/embeds/backup-locations.md
yaeltzirulnikov: yes, each backup file has a timestamp.

PR #378, opened 2019-05-16
Fixes to avoid issues in Hugo upgrade

bmansheim:


bmansheim:
@HarunD I see you were working on the upgrade. If you don't need this I'll close it.

HarunD:
@bmansheim yeah, you can close this one since I already am working on it.

PR #383, opened 2019-05-21
Fix TLS configuration images

bmansheim:


bmansheim:
@bmansheim /images/rs/crdb-tls-all.png is showing "TLS for CRDB communication only"

bmansheim:
Resolved in another PR

PR #386, opened 2019-05-23
Help user find the Import button in RC Pro

efratatredislabs:
pls add the location where you import a DB in Pro app. Toda


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/817

PR #387, opened 2019-05-23
Add explanation of RC Pro to FAQ

efratatredislabs:
lets clear the naming here....


bmansheim:
@Aviad-Abutbul Do you want to recommend a better explanation or just not to add the FAQ?

Issue #382, opened 2019-05-21
Improve display of search results in landing page


bmansheim:
1. The search dropdown should be the same size as the search box.
2. The search dropdown should show about 5 search results with the rest in the scroll. No need to have it drop past the bottom of the page.
3. Also, we might need to tweak the sizes of the icons so that they are all the same size as the Cloud Pro icon. How much work would that be?

https://user-images.githubusercontent.com/19514688/57847140-6b690300-77de-11e9-9325-af7afa0580d5.png
https://user-images.githubusercontent.com/19514688/57847311-d1558a80-77de-11e9-8d86-e4cdbec459a3.png

HarunD:
@bmansheim I fixed items 2. and 3.  I still need to reproduce the dropdown width issue on Windows so I can fix item 1.

PR #392, opened 2019-05-28
Landing page search tweaks

HarunD:
This PR fixes items from #382


bmansheim:
@HarunD Looks good. Can you fix the boxes in 1024 width so that there are even rows?

image

HarunD:
@bmansheim layout has been tweaked, rows are even now.

PR #395, opened 2019-05-30
Landing page icon tweaks

HarunD:


HarunD:
> It looks to me that the boxes are a pixel off from being aligned. Do you see that?

You're right, I tweaked the alignment that was of by a pixel.

PR #398, opened 2019-06-03
Track search via GA events

HarunD:
GA enabled on staging for testing purposes. Will disable it before merge.


bmansheim:
@HarunD Just tried the staging site and the search didn't bring up any results. Is it just me?

HarunD:
@bmansheim I did some changes to event tracking this morning and several search queries. As I do not have access to GA console we can wait until tomorrow afternoon to see whether these new events have been recorded.

PR #418, opened 2019-06-13
Hugo upgrade fixes with fix for expand

bmansheim:
* Use Permalink instead of URL

* Remove usage of .URL from menu

* Wrap .File usage in if check

* Update circleci config with new Hugo version


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/420

PR #419, opened 2019-06-14
kubectl get deployment command highlighted

vikrammoule-redislabs:
`kubectl get deployment`  goes unnoticed as it is not highlighted properly.


vikrammoule-redislabs:
@bmansheim I think we should also do something about the Copy to clipboard button for the commands. 
Along with the command, it copies the $ sign as well that you put to indicate shell. This forces the user to edit the command before hitting enter which defeats the purpose of having that button.

bmansheim:
@vikrammoule-redislabs Thanks for the insights. I'll take out the prompt.

PR #433, opened 2019-06-25
Add UNLINK to list of allowed multi-key commands

laurentdroin:


bmansheim:
This will be resolved in: https://github.com/RedisLabs/redislabs-docs/pull/431

PR #437, opened 2019-06-27
Fix display of cluster recovery steps (numbering)

vikrammoule-redislabs:
The step numbering is broken under Recovering the Cluster. So I manually put the numbers rather than depend on the auto-numbering.


bmansheim:
@vikrammoule-redislabs Thanks for pointing this out. There are more problems with the page that I need to address. Hopefully next week.

bmansheim:
Same problem on /rs/administering/database-operations/creating-database/

bmansheim:
Implementing changes in https://github.com/RedisLabs/redislabs-docs/pull/444

PR #446, opened 2019-07-04
Validate that GA is setup before using it

HarunD:
This change adds a check to see if GA is set up before trying to use it for tracking events like search.


bmansheim:
Tracking searches through GA site search

PR #448, opened 2019-07-08
Revert "Track search via GA events" but leave site search parameter

bmansheim:
Reverts RedisLabs/redislabs-docs#398


bmansheim:
We will track the search through Site Search of GA

PR #451, opened 2019-07-11
RED-35289 Unify RC Pro and Cloud docs

bmansheim:


bmansheim:
@HarunD For this PR we need to add a URL replace rule that replaces all of these URL requests:

- /rc/ > /rcessentials/
- /rv/ > /rcpro/

That way we won't have to make aliases for all of these pages.

bmansheim:
Here is the result:
image

bmansheim:
@HarunD I don't see that the rewrite rule in the htaccess file is working either on the staging site or on local build. Maybe we can do the redirect in the head.html file like you did with the /embeds ?

HarunD:
@bmansheim I will try that approach later today and let you know how it works.

PR #454, opened 2019-07-15
fix maintenance_mode flag names.

RoeyPrat:
do not backport to 5.4, as these flags were not updated in 5.4.2.


bmansheim:
@Paz-Yanover In keeping with the decision that our docs reflect the latest release and that we expect customers to always upgrade to the latest release, I will not add a note that in 5.4-5.4.4 these flags were different.

Paz-Yanover:
@bmansheim - right. The flags were changed in 5.4.6 (- and then we need to update the doc only after 5.4.6 is released) or in 5.4.4 (- and then we need to update the doc now)?

PR #456, opened 2019-07-16
RED-26867 Clearly mark internal ports

Paz-Yanover:
@bmansheim - I remember that we discussed this line with Yossi-
"| TCP | 9081 | For CRDB management |"
I understand that this port is used internally (in the node) and it should not be exposed outside the node. If it's correct, why did we decide to add it to this list? Do you remember?

### DIFF
--- a/content/rs/administering/designing-production/networking/port-configurations.md
+++ b/content/rs/administering/designing-production/networking/port-configurations.md
@@ -20,17 +20,17 @@ update your firewall with the port for that new database endpoint.
 | Protocol | Port | Description |
 |------------|-----------------|-----------------|
 | ICMP | * | For connectivity checking between nodes |
-| TCP | 1968 | Internal proxy usage. Ports are bound to loopback adapter. |
-| TCP | 3333, 3334, 3335, 3336, 3337, 3338, 3339, 36379, 36380 | Internal cluster usage |
-| TCP | 8001 | For application to access the RS [Discovery Service]({{< relref "/rs/concepts/data-access/discovery-service.md" >}}) |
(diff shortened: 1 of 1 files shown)


RoeyPrat:
relevant to ticket RED-26867

content/rs/administering/designing-production/networking/port-configurations.md
```diff
@@ -20,17 +20,17 @@ update your firewall with the port for that new database endpoint.
 | Protocol | Port | Description |
 |------------|-----------------|-----------------|
 | ICMP | * | For connectivity checking between nodes |
-| TCP | 1968 | Internal proxy usage. Ports are bound to loopback adapter. |
-| TCP | 3333, 3334, 3335, 3336, 3337, 3338, 3339, 36379, 36380 | Internal cluster usage |
-| TCP | 8001 | For application to access the RS [Discovery Service]({{< relref "/rs/concepts/data-access/discovery-service.md" >}}) |
-| TCP | 8443 | For secure (https) access to the management web UI |
-| TCP | 8444, 9080 | For nginx \<-> cnm_http/cm communications on the same host only. Ports are bound to loopback adapter. |
-| TCP | 9081 | For CRDB management |
+| TCP | 1968 | Proxy traffic (Internal use) |
```
RoeyPrat: @bmansheim my original intent was to have a separate column to mark internal use, or even a separate table, but whatever works.

bmansheim: If we could sort the table I would do that. Someday maybe.

PR #465, opened 2019-07-18
Page Feedback

HarunD:


bmansheim:
@HarunD The design comments I got on this were that when you click the down arrow, it should look like this:

Documentation_banner_Close

Also, ideally it would remember the state between pages so that if you minimize the feedback on one page and then go to another page it will also be minimized.

bmansheim:
@HarunD I also clicked on the feedback "Yes" in the staging site but I don't see any events in GA. Do we need to do something more to get it to show up?

bmansheim:
Here's the resulting report:
[Analytics docs.redislabs.com Page feedback 20190806-20190812 (1).pdf](https://github.com/RedisLabs/redislabs-docs/files/3495906/Analytics.docs.redislabs.com.Page.feedback.20190806-20190812.1.pdf)

PR #471, opened 2019-07-30
RED-30157 Add differences in dbsize and expires for crdb

bmansheim:


bmansheim:
@Paz-Yanover This is still waiting also.

bmansheim:
@RoeyPrat Can you verify this change?

content/rs/developing/crdbs/_index.md
```diff
@@ -191,6 +191,13 @@ the syncer initiates full reconciliation with other peers in the CRDB.
 
 ## INFO
 
+The INFO command shows these statistics differently for CRDBs:
+
+- DBSIZE (in `shard-cli info`) reports key header instances, not unique keys.
```
guybe7: DBSIZE is a command, it's not in the info string
maybe it's worth explaining what are key headers:
if replica1 did 'SET  x 1` and replica2 did `SADD x 1' simultaneously (i.e. connection between them was dropped) and then connection is restored, DBSIZE will show 2 but x is visibly a string on all replicas (in case of type conflicts such as that, string wins set)

content/rs/developing/crdbs/_index.md
```diff
@@ -191,6 +191,13 @@ the syncer initiates full reconciliation with other peers in the CRDB.
...
+- The keys count in DBSIZE (in `shard-cli info`) can be less than Expires (in `bdb-cli info`)
```
guybe7: i would phrase it the other way around: expires can be greater than DBSIZE

content/rs/developing/crdbs/_index.md
```diff
@@ -191,6 +191,13 @@ the syncer initiates full reconciliation with other peers in the CRDB.
...
+  because expires are not always removed when a key becomes a tombstone.
```
guybe7: maybe explain what is a tombstone? (a key that is logically deleted but still takes memory until it can be collected by the garbage collector)

content/rs/developing/crdbs/_index.md
bmansheim: How about: DBSIZE (in `shard-cli dbsize`) reports key header instances that represent multiple potential values of a key before a replication conflict is resolved.

PR #472, opened 2019-08-05
Delete excess files (embeds, categories) in circleci

bmansheim:


bmansheim:
@HarunD I got an notification from Google that there are usability errors on the embeds pages. To avoid Google seeing these pages, the embeds and categories directories are deleted at build time.
I also reverted the changes that we made to hide those directories from users.

HarunD:
@bmansheim Looks good

PR #477, opened 2019-08-08
Typos in persistence topic

silv-eran:


bmansheim:
@silv-eran Great typo fixing, Eran! For next time, our numbering is done automatically so no need to change the 2 to a 1. Otherwise, keep 'em coming.

PR #409, opened 2019-06-06
Update module names (RedisBloom and RedisJSON)

bmansheim:
https://docs.redislabs.com/staging/update-module-names/rs/getting-started/quick-setup/

PR #479, opened 2019-08-08
Update and rename rejson-quick-start.md to RedisJSON-quick-start.md

silv-eran:


bmansheim:
Addressed in https://github.com/RedisLabs/redislabs-docs/pull/409

PR #493, opened 2019-09-02
GCP support 

Aviad-Abutbul:


bmansheim:
@Aviad-Abutbul How does it look now?
https://docs.redislabs.com/staging/gcp-direct/rv/administration/setup_and_editing/view-edit-subscription/

PR #495, opened 2019-09-02
RED-31976 cluster recovery on k8s

otnielvh:


content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+---
+
+## Cluster Recovery
+If the cluster losses a majority of its nodes (e.g. from nodes crashing or network split), the cluster will not responde. Recovering the cluster requires performing the following steps:
+
+1. Edit the rec resource, and set the clusterRecovery flag to true, for example by running
+
+> kubectl patch rec <cluster-name> --type merge --patch '{"spec":{"clusterRecovery":true}}'
+
+2. Wait for the cluster to recover, i.e. wait for all the pods to reach a Running status. Tip, to see the progress of the cluster run:
```
danni-m: I would suggest here to take watch the progress via the `status` section

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+
+> watch kubectl get pod
+
+3. Recover the data by running the following command from within any cluster pod
```
danni-m: we need to add the full flow kubectl exec and then rladmin

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+
+> rladmin recover all 
+
+### Notes:
+
+1. For cluster recovery to work, we must be using persistent disks, otherwise there is no way to recvoer the cluster.
+
+2. If you want to restore a cluster from any ccs other than pod-0, you must copy the rdb file to pod-0 and recover from there.
+
+3. If you are using sentinel, you must restart the sentinel_service on the master by logging into the maste pod and running:
```
danni-m: @bmansheim @amiramm REF: RED-32898

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+description: 
+weight: 40
+alwaysopen: false
+categories: ["Platforms"]
+aliases: /rs/concepts/kubernetes/cluster-recovery
+---
+
+## Cluster Recovery
+If the cluster losses a majority of its nodes (e.g. from nodes crashing or network split), the cluster will not responde. Recovering the cluster requires performing the following steps:
+
```
amiramm: Couldn't find one but it would be preferable If we can refer to an article that explains how to detect a failure scenario that **mandates** recovery. We can give an example from `rladmin status` or of a split cluster.
We should also mention that this process can be used to intentionally tear down a cluster and rebuild it - for example for the purpose of recycling nodes, active/cold-passive failover, etc.

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+1. Edit the rec resource, and set the clusterRecovery flag to true, for example by running
+
+> kubectl patch rec <cluster-name> --type merge --patch '{"spec":{"clusterRecovery":true}}'
+
+2. Wait for the cluster to recover, i.e. wait for all the pods to reach a Running status. Tip, to see the progress of the cluster run:
+
+> watch kubectl get pod
+
```
amiramm: We need to explain how the cluster looks like after the recovery is complete - what is the expected state?

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+3. Recover the data by running the following command from within any cluster pod
+
+> rladmin recover all 
+
+### Notes:
+
+1. For cluster recovery to work, we must be using persistent disks, otherwise there is no way to recvoer the cluster.
+
```
amiramm: You can add a reference here to this article: https://docs.redislabs.com/latest/platforms/kubernetes/kubernetes-persistent-volumes/?s=persis
We've decided a while back to require persistence for production workloads, which is why it is worded as a requirement.

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+2. If you want to restore a cluster from any ccs other than pod-0, you must copy the rdb file to pod-0 and recover from there.
+
```
amiramm: nowhere in our public documentation today do we reference the ccs... we need to provide a little more context or generalize - for example, we can state that recovery in Redis Enterprise is performed from the master node and in Kubernetes we assume that pod <rec-name>-0 is the master node. In rare cases, when the master node is located in another pod and no replication has occurred for a prolonged period of time, it may be necessary to recover from a master node located on a different pod ordinal number. In order to accomplish that, the ccs-redis.rdb file from the /var/opt/redislabs/persist/ccs directory of the master node needs to be copied to the same directory in pod "-0". (what about the other rdb files? what about the difference in file format between AOF and sanpshot? do we need to mention that?)

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+3. If you are using sentinel, you must restart the sentinel_service on the master by logging into the maste pod and running:
+
+> supervisorctl restart sentinel_service
```
amiramm: is this performed as the last step of recovery?

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
...
+1. Edit the rec resource, and set the clusterRecovery flag to true, for example by running
+
+> kubectl patch rec <cluster-name> --type merge --patch '{"spec":{"clusterRecovery":true}}'
+
+2. Wait for the cluster to recover, i.e. wait for all the pods to reach a Running status. Tip, to see the progress of the cluster run:
+
+> watch kubectl get pod
+
+3. Recover the data by running the following command from within any cluster pod
+
```
amiramm: shouldn't we have another step here to run `rladmin status` to make sure we're in recover mode? `rladmin recover list` to make sure we have all the bdbs we need?

content/platforms/kubernetes/cluster-recovery.md
```diff
@@ -0,0 +1,33 @@
+---
```
amiramm: first and foremost - thanks for putting this together!

content/platforms/kubernetes/cluster-recovery.md
otnielvh: @danni-m Something like: `kubeclt exec -it <pod-name>` "raldmin recover all"

content/platforms/kubernetes/cluster-recovery.md
otnielvh: @amiramm , the other ccs-redis.rdb files don't need to be copied, the recovery is performed from a single ccs-redis.rdb file that resides on pod-0.
@danni-m , could the ccs-redis.rdb be an aof format? if yes do we need to mention it?

content/platforms/kubernetes/cluster-recovery.md
otnielvh: @amiramm , after the databases are recovered, the networking connectivity is wrong (for those using sentinel). Restarting sentinel solves this problem (it is a workaround for a bug)

content/platforms/kubernetes/cluster-recovery.md
bmansheim: @amiramm I don't think the other recovery article that we have adds much to this. It also might confuse readers because it is specifically in the context of RS.
I'll add the points about recycling nodes and active/passive failover.

content/platforms/kubernetes/cluster-recovery.md
bmansheim: Added the link

content/platforms/kubernetes/cluster-recovery.md
bmansheim: Improved reference to ccs

content/platforms/kubernetes/cluster-recovery.md
amiramm: Ideally, I would list which of the action in the RS article the operator does automate vs. the ones that still need to be done by the user, which are listed as steps in this article:
Operator automates all of these steps:

1. Recreates a fresh Redis Enterprise cluster
2. Mounts the persistent storage with the recovery files from the original cluster to the nodes of the new cluster
3. Recovers the cluster configuration on the first node in the new cluster
4. Joins the remaining nodes to the new cluster.

Agree though that it may confuse, so I'd leave it at your discretion.

content/platforms/kubernetes/cluster-recovery.md
danni-m: @otnielvh you can use bash and then rladmin, maybe its better if you plan to have more than 1 step with rladmin.
note that if you use a one liner you should use the full path for rladmin otherwise it will not work.

content/platforms/kubernetes/cluster-recovery.md
danni-m: >@danni-m , could the ccs-redis.rdb be an aof format? if yes do we need to mention it?
@otnielvh rdb file can hold both AOF and RDB formats, we don't need to mention it.

content/platforms/kubernetes/cluster-recovery.md
danni-m: @amiramm we can add this step but we need to think whats the actionable information the user is getting out of it.

amiramm:
> Thanks for putting this together @otnielvh !
> Please review my comments and address.
> @bmansheim - we need to time the release of this article with the version release.
@bmansheim release is now planned for September 30th.

amiramm:
@danni-m @otnielvh should we make a comment about persistent storage reclaim policies in the context of the persistence note? Advise to use `retain`?

danni-m:
@amiramm definitely, they can do that by modifying their storageclass or creating a new storageclass that has retain policy, for example:
https://medium.com/faun/kubernetes-how-to-set-reclaimpolicy-for-persistentvolumeclaim-7eb7d002bb2e

content/platforms/kubernetes/cluster-recovery.md
bmansheim: Added

content/platforms/kubernetes/cluster-recovery.md
bmansheim: Leaving it for now.

content/platforms/kubernetes/cluster-recovery.md
bmansheim: @danni-m Anything to add to the article about sentinel not updating endpoint?

content/platforms/kubernetes/cluster-recovery.md
bmansheim: Update the command according to Danni's comment?

amiramm:
@bmansheim we're releasing early. Please publish it as soon as you can.

PR #497, opened 2019-09-04
Secure ports usage

IAmATeaPot418:
@bmansheim & @Paz-Yanover - I struggled today to implement CRDB but ended up finding my way through it. I have some ideas on how we can improve how this is communicated but let's talk about those in person.
 
Updates below:
1. Updated to include the secure ports for use with CRDB instead of the insecure ports to help prevent insecure configurations
2. Updated to include instructions that would help fix a broken test case.

Thoughts?

### DIFF
--- a/content/rs/getting-started/getting-started-crdbs.md
+++ b/content/rs/getting-started/getting-started-crdbs.md
@@ -87,8 +87,8 @@ Now we have two Redis Enterprise Software clusters with FQDNs
     1. For the **database name**, enter: `database1`
     1. For the **endpoint port number**, enter: `12000`
     1. In the **participating clusters** list, add the address and admin credentials for:
-        - `http://cluster1.local:8080` - the cluster you are currently connected to
-        - `http://cluster2.local:8080` - the other cluster
+        - `https://cluster1.local:9443` - the cluster you are currently connected to
+        - `https://cluster2.local:9445` - the other cluster
     <!-- Also in create-crdb.md -->
     1. In the **Database clustering** option, either:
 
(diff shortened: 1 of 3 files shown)


content/rs/getting-started/getting-started-crdbs.md
```diff
@@ -75,6 +75,8 @@ Repeat the same operations for cluster 2 with these differences:
 Now we have two Redis Enterprise Software clusters with FQDNs
 **cluster1.local** and **cluster2.local**.
 
+Important Note: The cluster must be able to fully resolve DNS of each of the clusters and the databases that it is connecting to. Without this resolution the cluster will fail to connect successfully.
```
Paz-Yanover: I didn't get the "databases" point. Why the databases are relevant here?

IAmATeaPot418: for instance one of the things I had a problem with was ensuring that not only the cluster name resolved, but also that the DB name resolved. For instance it's not sufficient that cluster.local resolves but that redis-12000.cluster.local resolves. Otherwise, connections will fail.

Paz-Yanover: The cluster resolving is the responsibility of the customer, the db resolution is kind of ours...

IAmATeaPot418: @Paz-Yanover - I guess I'm confused if this is a bug or not. But that said the primary objective was the document the ports to use the encrypted port. Should we remove this section and move to change the port numbers?

Paz-Yanover: Sounds good to me

IAmATeaPot418: Just removed. @Paz-Yanover @bmansheim  anything else from your perspective?

Paz-Yanover: Fine with me

PR #499, opened 2019-09-07
Add shortcode for displaying tables based on CSV files

HarunD:


bmansheim:
@HarunD Looks good. Please update the css to match the regular table.

PR #507, opened 2019-09-14
Complete example for 3 nodes

shahar-biron:
Add 3rd node to the example of DNS configuration.


bmansheim:
@Paz-Yanover Any comments?

bmansheim:
Approved by @Paz-Yanover

PR #514, opened 2019-09-18
Added note about cluster recovery checkbox

amiramm:
Warning users not to check the recovery checkbox unless conducting cluster recovery.


amiramm:
need to update on Pivnet as well...

bmansheim:
@amiramm Is this text good?

amiramm:
> @amiramm Is this text good?

This isn't a strong enough warning. It should be clear that during normal installation this must be unchecked.

bmansheim:
Changed to: Make sure that the **Cluster Recovery** option is not selected during installation. This option is only for when you recover a cluster.

PR #515, opened 2019-09-18
Remove unnecessary reference

kerneljake:
Remove reference to section below; it is on a different page.


kerneljake:
This issue is also present on the [parent page](https://docs.redislabs.com/latest/rs/installing-upgrading/configuring/), but I do not know how to change the parent.

PR #516, opened 2019-09-18
Minor update to reflect changes in RCP

IAmATeaPot418:
Added that we have the ability to use append-only files every write. This is already in production.


IAmATeaPot418:
@Aviad-Abutbul @bmansheim - this is a pretty minor fix to document that we have the AOF fsync with every write capability in RCP.

bmansheim:
@Aviad-Abutbul AoF every 1 second is supported in RCP?

Aviad-Abutbul:
yes

PR #522, opened 2019-09-25
Update create-crdb.md

kamran-redis:
http  port mentioned in docs


IAmATeaPot418:
@kamran-redis  - this looks like it was originally correct. The instructions are describing how to navigate the web UI and not how to connect CRDB using the API. Am I missing something?

kamran-redis:
You are right I misread it

PR #530, opened 2019-10-08
Update getting-started-crdbs.md

WonkyWumpus:
The creation of the CRDB database is executed locally on cluster1.local and uses the docker network to connect to cluster2.local.  The API is available on 9443, not 9445.  The API is mapped to 9445 on the host machine only.  Attempting to connect with the 9445 address gave me a connection refused error. 9443 completes successfully.


bmansheim:
@WonkyWumpus Thanks for catching that!

PR #535, opened 2019-10-10
Add lfu eviction policies for RC Pro

ruslan-redislabs:
Missing recently added *-lfu eviction policies

### DIFF
--- a/content/rv/concepts/data-eviction-policies.md
+++ b/content/rv/concepts/data-eviction-policies.md
@@ -11,8 +11,10 @@ policies:
 |  **Options** | **Description** |
 |------------|-----------------|
 |  allkeys-lru | Evicts the least recently used (LRU) keys out of all keys in the database |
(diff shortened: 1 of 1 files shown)


content/rv/concepts/data-eviction-policies.md
```diff
@@ -11,8 +11,10 @@ policies:
 |  **Options** | **Description** |
 |------------|-----------------|
 |  allkeys-lru | Evicts the least recently used (LRU) keys out of all keys in the database |
+|  allkeys-lfu | Evicts the least frequently used keys out of all keys
```
ushachar: We should prob. match with the descriptions in https://redis.io/topics/lru-cache (maybe change there as well?)

joniredis: In my "new to redis" opinion, I think that the description on redis.io properly matches the more in depth explanation about the approximization and would leave it there as is. For RCP I would use the more simplified description that is suggested here.

PR #536, opened 2019-10-10
Remove note to receive notifications on node join and remove (alerting)

ofirluzon:
removed cluster alerts note from database alert doc


ofirluzon:
See #537

PR #537, opened 2019-10-10
Add note to receive notifications on node join and remove (alerts)

ofirluzon:
Add the note in the relevant page


ofirluzon:
see #536

PR #539, opened 2019-10-18
Local user password policy

IAmATeaPot418:
@Paz-Yanover @bmansheim - I've included the below 3 modifications in this PR.

1. Updated our API call to add external users to leverage our secure API and not the unencrypted one.
2. Documented the local password complexity profile we have.
3. Documented the local password expiration profile we have.


bmansheim:
@Paz-Yanover @IAmATeaPot418 I made some style and format changes and also added password reset instructions. Unless you object, I'll merge it tomorrow.

bmansheim:
https://docs.redislabs.com/staging/Local-User-Password-Policy/rs/administering/designing-production/security/account-management/

PR #545, opened 2019-10-23
New header and footer style

bmansheim:


bmansheim:
@HarunD I added a few tweaks for aesthetics. I don't see the menus opening though. Is it just me?

PR #547, opened 2019-10-28
RED-33986 Improve edit CRDB

bmansheim:


content/rs/administering/database-operations/updating-configurations.md
```diff
@@ -36,6 +36,8 @@ most database settings only apply to the CRDB instance that you are editing, inc
 - TLS mode
 - Periodic backup
 
+You can change the global configuration of the CRDB from the command-line with: `crdb-cli crdb update`
```
bmansheim: @Paz-Yanover Do we want to reference this command here? It seems important for customers to know how to update the global CRDB configuration so I think it needs to be here.

Paz-Yanover: Maybe we should just mention that the global configuration can be changed using crdb-cli and link to the crdb-cli documentation (- let's add it)?

PR #550, opened 2019-10-30
RedisInsight k8s support

K-Jo:
WIP


amiramm:
I suggest we store this yaml in a public GitHub repo or a website.
We can still list it here but instead of the operator needing to copy-paste this into a file and potentially make an indentation or other errors he/she would be able to download it (for an air-gapped environment) or apply it directly from the Web, like in this example:
kubectl apply -f https://k8s.io/examples/application/deployment.yaml

PR #551, opened 2019-10-31
improve description of accumulated latency

oranagra:
the same change should be applied on all acc_latency metrics (read/write/other).
and should also be put in other docs we have about this metrics (rest api?).
also note that the acc_latency_max metric is completely useless IMHO, since the second with the highest latency is not necessarily the one with the highest response time, so any division attempt to get the real latency would be wrong.
not sure if we should mention that it is useless, or trim it from the docs.


Paz-Yanover:
@bmansheim - per Oran's comment, the same change should be applied to other metrics as well

bmansheim:
@Paz-Yanover Please approve.

PR #553, opened 2019-11-04
Update compatibility with OSS cluster

bmansheim:


bmansheim:
@Paz-Yanover @AlonMagrafta Do you guys what to approve?

PR #555, opened 2019-11-04
Update multi-key limitation for OSS cluster API

ofirluzon:
Changed the multi-key disclaimer as the former was very wrong.


bmansheim:
@ofirluzon Please review my commit

PR #559, opened 2019-11-07
Warn not to do global operations before upgrading all CRDBs

Paz-Yanover:
valentino@redislabs.com, @bmansheim - please review and approve

### DIFF
--- a/content/rs/installing-upgrading/upgrading.md
+++ b/content/rs/installing-upgrading/upgrading.md
@@ -130,8 +130,8 @@ instances that were not upgraded yet cannot receive write updates from the upgra
 The upgraded instance receives updates from upgraded and non-upgraded instances.
 
 {{% note %}}
-- Upgrade all instances of a specific CRDB within a reasonable time frame to avoid temporary inconsistencies between the instances.
-- Make sure that you upgrade all instances of a specific CRDB before you do global operations on the CRDB such as removing instances and adding new instances.
+- Upgrade all instances of a specific CRDB within a reasonable time frame.
+- Make sure that you upgrade all instances of a specific CRDB before you do global operations on the CRDB, such as removing instances and adding new instances.
 {{% /note %}}
 
 After you upgrade an instance to use the new protocol version,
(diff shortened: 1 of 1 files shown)


bmansheim:
@Paz-Yanover Please review my commit

bmansheim:
@Paz-Yanover Can you make an alternative suggestion?

Paz-Yanover:
@bmansheim - see https://github.com/RedisLabs/redislabs-docs/pull/559/commits/3c006a4b6e7c4fff29b359c676e8b7305736f72d

bmansheim:
@Paz-Yanover We want to have the note higher up on the page to make sure customers see it, and the two ideas are similar. How does it look now?

Paz-Yanover:
@bmansheim - I don't think the sentence is correct, as the second point is not "to avoid temporary inconsistencies between the instances".
@valentino@redislabs.com - what do you think?

Paz-Yanover:
@bmansheim - Valentino agrees with my comment above

Paz-Yanover:
@bmansheim -
Let's add the following:
{{% note %}}
Until upgrading all instances of a specific CRDB, avoid performing global operations on this CRDB, like removing instance, adding new instance etc.
{{% /note %}}

content/rs/installing-upgrading/upgrading.md
```diff
@@ -130,8 +130,8 @@ instances that were not upgraded yet cannot receive write updates from the upgra
 The upgraded instance receives updates from upgraded and non-upgraded instances.
 
 {{% note %}}
-- Upgrade all instances of a specific CRDB within a reasonable time frame to avoid temporary inconsistencies between the instances.
-- Make sure that you upgrade all instances of a specific CRDB before you do global operations on the CRDB such as removing instances and adding new instances.
+- Upgrade all instances of a specific CRDB within a reasonable time frame.
```
Paz-Yanover: @bmansheim - why did you remove the "reason" here?

PR #521, opened 2019-09-25
RED-32998 Add Modules section with all related content

bmansheim:

PR #563, opened 2019-11-12
Note on eviction policy

ruslan-redislabs:
Eviction policy set to "noeviction" with modules enabled. Added a "Note:" on line 39.

### DIFF
--- a/content/rs/developing/modules/_index.md
+++ b/content/rs/developing/modules/_index.md
@@ -36,6 +36,7 @@ Module]({{< relref "/rs/developing/modules/upgrading.md" >}})
 
 Warning: Redis Labs does not and cannot support third party modules or
 databases created with them.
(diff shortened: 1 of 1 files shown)


maguec:
:+1:

K-Jo:
I don't believe this is a correct modification to our docs.  Each module declares it's enterprise capabilities.  We should find a way to explain those in our documenation
https://github.com/RedisLabs/RAMP#module-capabilities

ruslan-redislabs:
Maybe a matrix per module/capability and a reference to it? Or in general, "If using modules, consider checking modules capability to see if eviction policies are supported"?

bmansheim:
Resolved in https://github.com/RedisLabs/redislabs-docs/pull/521/commits/bc757936699a650cc002797f9b6b2333eb39c15a

PR #568, opened 2019-11-15
Changed the title of the IOPS panel

laurentdroin:
IOPS is confusing. This panel is only applicable for RedisOnFlash and relates to IOs. Customers who don't see any data and who expect the metrics to show instantaneous operations per second are wondering why the graph as no data. Changing the title to indicate that 1) it is for RoF, and 2) it is disk I/Os seems like a good change.

### DIFF
--- a/static/code/rs/database.json
+++ b/static/code/rs/database.json
@@ -1865,7 +1865,7 @@
         "thresholds": [],
         "timeFrom": null,
         "timeShift": null,
-        "title": "IOPS",
+        "title": "Flash disk IOs",
         "tooltip": {
             "shared": true,
             "sort": 0,


danni-m:
@laurentdroin @bmansheim im not sure I understand, what am I looking at?

laurentdroin:
> @laurentdroin @bmansheim im not sure I understand, what am I looking at?

@danni-m It is about this page: https://docs.redislabs.com/latest/rs/administering/monitoring-metrics/prometheus-integration/

The proposed modification is for step 4, paragraph c. It is the JSON for the proposed database dashboard. The title of the IOPS panel is misleading customers so we would like to clarify it.

danni-m:
LGTM

PR #569, opened 2019-11-15
Update installing RedisInsight desktop for Linux

svntjng:
Add a bit more details about how to make the downloaded file executable and how to execute them on Linux.


bmansheim:
@svntjng Thanks for contributing!

PR #571, opened 2019-11-20
Cloud api red 35259 add vpc peering

dannycohen:


bmansheim:
Resolved in https://github.com/RedisLabs/redislabs-docs/pull/572

PR #573, opened 2019-11-25
RED-35188 Add caveat for non-identical ccs files

bmansheim:


content/rs/administering/troubleshooting/cluster-recovery.md
```diff
@@ -79,7 +79,8 @@ of the configuration and persistence files on each of the nodes.
     {{% expand "Command syntax" %}}
 `<filename>` - The full path of the old cluster configuration file in the persistent storage.
 The cluster configuration file is /css/ccs-redis.rdb.
-The file is located on the persistent storage drive for each node in the original cluster and all copies are identical.
+Before you copy the ccs file, confirm that all of the ccs files on each node are identical.
+If the ccs files are not identical, contact Redis Labs Support for assistance.
```
galcohen-redislabs: They will never be identical, so this comment will confuse and cause an unnecessary noise.

bmansheim:
@Paz-Yanover It seems to me that we need to make some change to this because the current text "all copies are identical" is incorrect. What do you suggest?

Paz-Yanover:
@bmansheim - since we don't have a clever suggestion for the customers, maybe we should just leave it as is.

bmansheim:
@Paz-Yanover Added it to the prerequisites. Please approve.

content/rs/administering/troubleshooting/cluster-recovery.md
```diff
@@ -40,6 +40,9 @@ The cluster recovery process includes:
 - We recommend that you use clean persistent storage drives for the new cluster.
     If you use the original storage drives,
     make sure that you backup the files on the original storage drives to a safe location.
+- The cluster configuration file that you want to use as the configuration for the cluster.
+    The cluster configuration file is `/css/ccs-redis.rdb` on each of the nodes.
```
Paz-Yanover: This paragraph is not accurate (i.e. this sentence and the next one), as the file is located on the persistent storage drive for each node. It's not necessarily on the node itself.

bmansheim: Change to; The cluster configuration file is `/css/ccs-redis.rdb` on the persistent storage drive for each node. OK?

content/rs/administering/troubleshooting/cluster-recovery.md
```diff
@@ -40,6 +40,9 @@ The cluster recovery process includes:
 - We recommend that you use clean persistent storage drives for the new cluster.
     If you use the original storage drives,
     make sure that you backup the files on the original storage drives to a safe location.
+- The cluster configuration file that you want to use as the configuration for the cluster.
+    The cluster configuration file is `/css/ccs-redis.rdb` on the persistent storage for each node.
+    Copy the cluster configuration file from one of the nodes to the persistent storage that you want to connect to the first node.
```
Paz-Yanover: As said before- this sentence is inaccurate. No need to copy anything, and the file is not located on the nodes...

PR #575, opened 2019-11-25
Added a bit of clarity to description in the beginning.

ruslan-redislabs:
Added reference to the cluster recovery process in lines 11-15 as customers are getting confused.


bmansheim:
@ruslan-redislabs Good point! I tried to make it a little more clear. What do you think?

Also, please sign the CLA so it doesn't block merges. Thanks!

ruslan-redislabs:
@bmansheim looks good and more structured. I've signed the CLA too, sorry haven't done it earlier )

PR #591, opened 2019-12-04
update create-crdb.md

TobyHFerguson:
Line 81 had a recursive and never ending definition of <URL>. I corrected this to show how the url was really formed


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/593

PR #594, opened 2019-12-05
Fix two RZ to three RZ as minimum for RZA

shahar-biron:


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/599

PR #610, opened 2019-12-11
Change "node" to "slave" where applicable

erezatiya:
Might as well remove "shard" in most of the paragraph and only mention it once.
The reasoning: the dataset is actually stored on the shards, and when a node fails which contains master shards, a failover happens.

A dataset is not actually replicated to a slave node, but a slave shard. A *node* failover only occurs if the master node fails. There is one master node in a cluster and the rest are slaves.


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/598

PR #612, opened 2019-12-13
RedisInsight: Added new section on configuring RedisInsight using environment variables.

nishantgeorge:
Screenshot from 2019-12-13 13-49-31


nishantgeorge:
Screenshot from 2019-12-16 17-07-32

nishantgeorge:
> The port being one of them. @nishantgeorge there was a second one we also needed to document, is it the binding port?

Yes, I think there was one but I can't remember which.

lhriley:
@nishantgeorge thanks for getting this started.

I've been looking for ways to configure RedisInsight via environment variables so I can deploy it in our k8s cluster. Are there any additional variables that we can manage? For example, by viewing the docker container environment I found `REDISINSIGHT_HOME_DIR`.

For instance, I would really like to be able to setup the connection to a running redis instance via environment variables if possible.

nishantgeorge:
> I've been looking for ways to configure RedisInsight via environment variables so I can deploy it in our k8s cluster. Are there any additional variables that we can manage? For example, by viewing the docker container environment I found `REDISINSIGHT_HOME_DIR`.

Thanks for pointing that out :) I've added documentation for that and other environment variables that are meant to be configured by end users.
 
> For instance, I would really like to be able to setup the connection to a running redis instance via environment variables if possible.

There is no provision for this at the moment, but there has been some discussion around making it easy to script adding databases. I'll defer this feature request to @K-Jo

PR #613, opened 2019-12-13
RedisInsight: Updated docs on connecting to ElastiCache.

nishantgeorge:
Screenshot from 2019-12-13 17-37-15

### DIFF
--- a/content/ri/using-redisinsight/add-instance.md
+++ b/content/ri/using-redisinsight/add-instance.md
@@ -31,7 +31,19 @@ Connecting to AWS ElastiCache can be particularly problematic since ElastiCache 
 
 If you want to work with ElastiCache Redis caches using RedisInsight, you have two options:
 
+1. If you are not using redis cluster, you can [setup an SSH Tunnel](https://userify.com/blog/howto-connect-redis-ec2-ssh-tunnel-elasticache/) between RedisInsight and your Elasticache instance. An SSH tunnel consists of an encrypted tunnel created through an SSH protocol connection.
+
+   - Run this command to create an ssh tunnel:
+
+     ```bash
+      ssh -f -N -L <local_port>:<elasticache_endpoint> -i <path_to_pem_file> <ec2_endpoint>
+      ```
+      
+   - Go to Add Instance in RedisInsight and add an instance with host=localhost, port=<local_port>,
+     name=your_instance_name
+
 1. [Install RedisInsight on an EC2 instance]({{< relref "/ri/installing/install-ec2.md" >}}) that's in the same VPC and has access to the ElastiCache Redis cache. 
-   This is the easiest option and yields the best performance.
+   This is option yields the best performance and works with Redis Cluster.
+
 1. [Set up a VPN to your AWS VPC using AWS VPN](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/accessing-elasticache.html#access-from-outside-aws). 
    You can then access the ElastiCache Redis cache using the private endpoint.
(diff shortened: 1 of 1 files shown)


content/ri/using-redisinsight/add-instance.md
```diff
@@ -27,23 +27,13 @@ Troubleshooting:
...
 ## Connecting to ElastiCache
 
-If you want to work with ElastiCache Redis instances with RedisInsight, you have two options:
+Connecting to AWS ElastiCache can be particularly problematic since ElastiCache caches cannot be accessed from outside the VPC, as they don't have public IP addresses assigned to them.
 
-1. [Install RedisInsight on an EC2 instance]({{< relref "/ri/installing/install-ec2.md" >}}) that has access to the ElastiCache Redis instance.
-1. Create a SSH tunnel through an EC2 instance that can access the ElastiCache Redis instance.
+If you want to work with ElastiCache Redis caches using RedisInsight, you have two options:
 
-### Using an SSH Tunnel
```
K-Jo: why did we delete this level of detail.  Is it no longer relevant?

content/ri/using-redisinsight/add-instance.md
```diff
@@ -27,23 +27,13 @@ Troubleshooting:
 
 ## Connecting to ElastiCache
 
-If you want to work with ElastiCache Redis instances with RedisInsight, you have two options:
+Connecting to AWS ElastiCache can be particularly problematic since ElastiCache caches cannot be accessed from outside the VPC, as they don't have public IP addresses assigned to them.
 
-1. [Install RedisInsight on an EC2 instance]({{< relref "/ri/installing/install-ec2.md" >}}) that has access to the ElastiCache Redis instance.
-1. Create a SSH tunnel through an EC2 instance that can access the ElastiCache Redis instance.
+If you want to work with ElastiCache Redis caches using RedisInsight, you have two options:
```
K-Jo: you have three options

content/ri/using-redisinsight/add-instance.md
```diff
@@ -27,23 +27,13 @@ Troubleshooting:
...
-    ```
-
-1. Go to Add Instance in RedisInsight and add an instance with host=localhost, port=8765,
-   name=your_instance_name
-
-1. You are now connected to your elasticache instance and can start using it locally.
+1. [Install RedisInsight on an EC2 instance]({{< relref "/ri/installing/install-ec2.md" >}}) that's in the same VPC and has access to the ElastiCache Redis instance. 
+   This is the easiest option and yields the best performance.
+1. [Use a NAT instance to forward traffic between your ElastiCache instance and RedisInsight.](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/accessing-elasticache.html#access-from-outside-aws) 
+   This option has a non-trivial setup, and is not recommended by AWS. Also, this *does not work with cluster-mode-enabled ElastiCache caches*.
```
K-Jo: needs link to not recommended by aws
needs link to why NAT + redis cluster

content/ri/using-redisinsight/add-instance.md
K-Jo: I would really keep this

nishantgeorge: I removed it because it is essentially a poor man's NAT. But it is definitely much easier to set up. Agreed that we should keep it.

bmansheim:
@nishantgeorge Please update according to the comment.

nishantgeorge:
@K-Jo @bmansheim I've updated the docs to align with AWS's updated docs, that have removed the recommendation for NAT instance and instead recommends and provides detailed steps for the AWS VPN option.
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/accessing-elasticache.html#access-from-outside-aws

PR #618, opened 2019-12-20
Added GCP as an option into quick-setup.md

TobyHFerguson:
Needs a new image showing the GCP option - I added a suitable screenshot here.
Screen Shot 2019-12-19 at 3 57 17 PM


bmansheim:
@TobyHFerguson Redis Cloud just got a big upgrade that makes this change obsolete. Thanks for reporting, though!

PR #623, opened 2019-12-24
minor grammatical edits and technical improvements

carehart:


bmansheim:
@carehart Thank you for your contribution. You're right that this article needs to be improved. We have another PR open to change this article. It should be updated soon.

PR #624, opened 2019-12-25
RED-35288 Simplified cloud signup

bmansheim:


bmansheim:
Simplified in https://github.com/RedisLabs/redislabs-docs/pull/625

PR #633, opened 2020-01-02
RED-34932 Support for RHEL7.7

bmansheim:


bmansheim:
Merged to 5.4.14 by mistake but reverted and resubmitted

PR #635, opened 2020-01-05
Add Connection Source column

shahar-biron:
Easier reference for customers which ports should be open by Firewall in either Geo distribution, or simple Client / Server access.

### DIFF
--- a/content/rs/administering/designing-production/networking/port-configurations.md
+++ b/content/rs/administering/designing-production/networking/port-configurations.md
@@ -20,17 +20,17 @@ update your firewall with the port for that new database endpoint.
 | Protocol | Port | Connection Source | Description |
 |------------|-----------------|-----------------|-----------------|
 | ICMP | * | Internal | For connectivity checking between nodes |
-| TCP | 1968 | Internal | Proxy traffic  |
-| TCP | 3333, 3334, 3335, 3336, 3337, 3338, 3339, 36379, 36380 | Internal | Cluster traffic  |
+| TCP | 1968 | Internal | Proxy traffic |
+| TCP | 3333, 3334, 3335, 3336, 3337, 3338, 3339, 36379, 36380 | Internal | Cluster traffic |
 | TCP | 8001 | Internal, External | Traffic from application to RS [Discovery Service]({{< relref "/rs/concepts/data-access/discovery-service.md" >}}) |
 | TCP | 8443 | Internal, External | Secure (HTTPS) access to the management web UI |
-| TCP | 8444, 9080 | Internal | For nginx <-> cnm_http/cm traffic  |
-| TCP | 9081 | Internal, Active-Active | For Active-Active management  |
+| TCP | 8444, 9080 | Internal | For nginx <-> cnm_http/cm traffic |
+| TCP | 9081 | Internal, Active-Active | For Active-Active management |
 | TCP | 8070, 8071 | Internal, External | For metrics exported and managed by nginx |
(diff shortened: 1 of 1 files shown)


content/rs/administering/designing-production/networking/port-configurations.md
```diff
@@ -17,20 +17,25 @@ update your firewall with the port for that new database endpoint.
...
-| TCP | 8001 | Traffic from application to RS [Discovery Service]({{< relref "/rs/concepts/data-access/discovery-service.md" >}}) |
-| TCP | 8443 | Secure (HTTPS) access to the management web UI |
-| TCP | 8444, 9080 | For nginx <-> cnm_http/cm traffic (Internal use) |
-| TCP | 9081 | For CRDB management (Internal use) |
-| TCP | 8070, 8071 | For metrics exported and managed by nginx |
-| TCP | 9443 (Recommended), [8080](#turning-off-http-support) | REST API traffic, including cluster management and node bootstrap |
-| TCP | 10000-19999 | Database traffic |
-| TCP | 20000-29999 | Database shards traffic (Internal use) |
-| UDP | 53, 5353 | DNS/mDNS traffic (Internal use) |
+| Protocol | Port | Connection Source | Description |
```
Paz-Yanover: I think "Use" is a better title... @bmansheim - what do you think?

content/rs/administering/designing-production/networking/port-configurations.md
```diff
@@ -17,20 +17,25 @@ update your firewall with the port for that new database endpoint.
...
+|------------|-----------------|-----------------|-----------------|
+| ICMP | * | Internal | For connectivity checking between nodes |
+| TCP | 1968 | Internal | Proxy traffic  |
+| TCP | 3333, 3334, 3335, 3336, 3337, 3338, 3339, 36379, 36380 | Internal | Cluster traffic  |
+| TCP | 8001 | Internal, External | Traffic from application to RS [Discovery Service]({{< relref "/rs/concepts/data-access/discovery-service.md" >}}) |
+| TCP | 8443 | Internal, External | Secure (HTTPS) access to the management web UI |
+| TCP | 8444, 9080 | Internal | For nginx <-> cnm_http/cm traffic  |
+| TCP | 9081 | Internal | For CRDB management  |
+| TCP | 8070, 8071 | Internal, External | For metrics exported and managed by nginx |
+| TCP | 9443 (Recommended), [8080](#turning-off-http-support) | Internal, External, Geo Distribution | REST API traffic, including cluster management and node bootstrap |
```
Paz-Yanover: Let's avoid the "Geo Distribution" in this column at all

shahar-biron: Without having "Geo Distribution" we will have hard time telling customers which ports they need to open between clusters... External is "too general" as it refers to local data center and not to a Remote data center is in Active/Active. If you prefer "Multi data Center" instead - so be it, but we need this category.

Paz-Yanover: We can add a comment, that this port is used by A-A as well, but I don't think we need to add it as a different / dedicated usage

content/rs/administering/designing-production/networking/port-configurations.md
bmansheim: @Paz-Yanover I actually like "Connection Source"

bmansheim:
@shahar-biron and @Paz-Yanover please review again

Paz-Yanover:
@bmansheim - again, I don't like this clear separation between internal, external and Active-Active

PR #638, opened 2020-01-12
expand on Storage Class

shahar-biron:
include Storage Classes provided by main 3 public cloud providers (added Azure)


amiramm:
Looks good, thanks @shahar-biron !

PR #649, opened 2020-01-20
Update private-public-endpoints.md

kerneljake:
Minor grammar fixes; also a suffix would not contain a node name.


galcohen-redislabs:
@kerneljake Best avatar in all of redislabs!

galcohen-redislabs:
@bmansheim It is approved, but for some reason github doesn't show it...

PR #652, opened 2020-01-23
Add k8s cluster recovery workaround to cluster recovery article

laurentdroin:
I think that the known limitation should be added to this page and moved out of the 5.4.10 release notes. This is not an issue that pertains to 5.4.10 and customers who are going to do a recovery are more likely to check the recovery documentation than the release notes.
I copied the text straight from the 5.4.10 release notes (and I left it there but I don't like duplicate information since it's hard to maintain. We should remove the next from the release notes IMO).

### DIFF
--- a/content/embeds/force-delete-pods.md
+++ b/content/embeds/force-delete-pods.md
@@ -1,8 +1,10 @@
 In some cases, pods do not terminate when the statefulSet is scaled down as part of the cluster recovery.
-If pods are stuck in `terminating` or `crashLoopBack` and do not terminate gracefully, cluster recovery can stop.
+If pods are stuck in `terminating` or `crashLoopBack` and do not terminate gracefully, cluster recovery can pause.
 
-In that case, delete the pods manually with:
+To work around this, delete the pods manually with:
 
 ```src
 kubectl delete pods <pod> --grace-period=0 --force
 ```
+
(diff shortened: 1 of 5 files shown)


bmansheim:
@laurentdroin Thanks for the contribution. I made some formatting changes and put the shared content in a shared file. Let me know if you have comments.

bmansheim:
Staging page at: https://docs.redislabs.com/staging/laurentdroin-patch-2/platforms/kubernetes/cluster-recovery/

laurentdroin:
@bmansheim  In "In some cases, pods do not terminate when the statefulSet is scaled down as part of the cluster recovery. If pods are stuck in terminating or crashLoopBack and do not terminate gracefully, cluster recovery can stop.", I would not say that the recovery could "stop". Stop may make people think that the recovery would be restarted, which is not the case. I guess we could say "pause". As soon as the last pod is manually delete, the recovery process resumes automatically without intervention

laurentdroin:
@bmansheim - Also, this is to be verified with Product.R&D, I don't think that having persistence on the database is a pre-requisite for cluster recovery. If you don't have database persistence enabled, the recovery will work, recreating the cluster and the databases. Databases that don't have persistence enabled will be recreated empty.

bmansheim:
@laurentdroin Thanks. I updated the PR according to your comments.

PR #655, opened 2020-01-24
Adding all latest ga module release notes

K-Jo:


K-Jo:
@bmansheim we still need release notes for Search and Graph in this PR let's do it at once

PR #663, opened 2020-02-04
Update redis-flash.md

kerneljake:
Product name change.


bmansheim:
@kerneljake Thanks for spotting that. I'll fix a few more that I found in the same PR.

PR #664, opened 2020-02-04
Add a link to the full installation

dallastexas92:
This is a great 'getting started' guide for a basic deployment of Redis Enterprise. As folks might start here and then jump to the full installation instructions (including the required ports, etc), it would be great if there were a link on this page to the full installation instructions (rather than having to find it).

Note: Don't need to take my wording EXACTLY but something along those lines would be beneficial IMO


bmansheim:
@dallastexas92 Thanks Dallas for the contribution. It gives me a good opportunity to make some other changes also.

PR #666, opened 2020-02-06
RS quick setup with embed and links to full install

bmansheim:


content/embeds/cluster-setup.md
```diff
@@ -0,0 +1,26 @@
...
+    ![Redis Enterprise Software Setup](/images/rs/getstarted-setup.png?width=600)
+
+1. In the **Node Configuration** settings, enter a cluster FQDN such as `cluster.local`.
+Then click **Next** button.
+
+    ![Redis Enterprise Software node configuration](/images/rs/getstarted-nodeconfig.png?width=600)
+
+1. Enter your license key, if you have one. If not, click the **Next** button to use the trial version.
+
+1. Enter an email and password for the admin account for the web console.
```
AlonMagrafta: Actually it's an Admin for the UI as well as for the cluster.
Meaning it's also the Admin user to send API command with its credentials

bmansheim: Added "These credentials are also for the connections to the REST API." OK?

AlonMagrafta: how about "These credentials are also to use the Redis REST API."?

PR #667, opened 2020-02-06
Update terminology to Active-Active

bmansheim:


bmansheim:
Resubmitted as: https://github.com/RedisLabs/redislabs-docs/pull/928

PR #668, opened 2020-02-06
RED-25970 CRDB flush UI

bmansheim:


bmansheim:
@AlonMagrafta I updated the terms to Active-Active

PR #670, opened 2020-02-06
Typo

alanmcg:
Fixed typo


bmansheim:
Thanks for catching that @alanmcg !

PR #686, opened 2020-02-19
Standardize Module and Platform RN titles

bmansheim:
@K-Jo If you approve these changes, I'll make them also on the other module RNs.


K-Jo:
@bmansheim if we standardise it, is it for all modules the case?  I find the argument of @AlonMagrafta unneeded. the folder structure already indicates it is the RN of RedisBloom

PR #687, opened 2020-02-20
RED-34932 Support for RHEL7.7

bmansheim:
Approved in https://github.com/RedisLabs/redislabs-docs/pull/633


AlonMagrafta:
@Paz-Yanover We said that from 5.6.0 (and stated in RN) we will base our Oracle support on the RHEL. I think we should edit it here as well. Agree?
Also extending the 5.4 EOL and stating RHEL6 will end at nov 30 this year

Paz-Yanover:
@AlonMagrafta - we can state here regarding Oracle. 
Regarding 5.4 EOL and RHEL6 EOL- do we need to state it here? until now we didn't state any EOL here.

AlonMagrafta:
@Paz-Yanover not here. 
5.4 - is already in the lifecycle page. We don't have a place for RHEL - probably since it is according to the support of Red hat for it

PR #693, opened 2020-02-26
RedisInsight docs: Updated instructions to add db

bmansheim:
Added detailed instructions with screenshots for adding different setups of Redis including:
- Standalone Redis server
- Redis Cluster OSS
- Redis Sentinel OSS
- Redis Enterprise with mutual TLS


nishantgeorge:
@bmansheim Where can I see the staging site?

content/ri/using-redisinsight/api.md
```diff
@@ -0,0 +1,178 @@
...
+**Auth required** : NO
+
+### Parameters
+
+These are the required parameters for any type of database.
+
+| Parameter     | Type   | Description         |
+|---------------|--------|---------------------|
+| name          | string | A nick name for the Redis database. Any string is valid |
+| dbType        | string | One of `"OSS_NON_CLUSTER"`, `"OSS_CLUSTER"` or `"OSS_SENTINEL"`. For any Redis Enterprise database, use `OSS_NON_CLUSTER` |
```
K-Jo: @nishantgeorge Why don't we have a redis enterprise type?  Btw we can have redis enterprise cluster as well.  Why wouldn't we add this so that special features can be detected.  We should even have them for software and cloud no?

content/ri/using-redisinsight/api.md
```diff
@@ -0,0 +1,178 @@
...
+|--------------------------|---------|---------------------------------------------------------------|
+| useTls                   | boolean | Whether to use TLS to connect to the database or not          |
+| clientAuth               | boolean | Whether TLS client authentication is required by the database |
+| clientCertificateKeyPair | object  | (optional) [The details of the client certificate and private key used to connect to the Redis database](#tls-client-certificate-and-key). If client authentication is not required, this has to be provided |
+
+##### TLS Client Certificate and Key
+
+The client certificate and key details can be provided in two forms:
+
+1. If the certificate and key has already been used for a database before, the ID of that certificate can be provided directly.
```
K-Jo: @nishantgeorge so how do you get the list of id's?  Shouldn't we document the call to get the certificates as well?

K-Jo: This section might also benefit from an example after we have that rest call documented

content/ri/using-redisinsight/api.md
```diff
@@ -0,0 +1,178 @@
...
+Title: RedisInsight REST API
+date: 2020-02-02 00:00:00 +0000
+weight: 1000
+categories: ["RI"]
+path: api/
+nextStep:
+    Title: Memory Analysis
+    href: /docs/features/memory-analysis/
+---
+
```
K-Jo: @nishantgeorge I'm missing a section here on why we did this.  We wanted this feature to be documented so that people could script / bootstrap a redisinsight with their databases preloaded for automation / testing purposes ?

content/ri/using-redisinsight/api.md
```diff
@@ -0,0 +1,178 @@
+---
+Title: RedisInsight REST API
```
K-Jo: Not sure if we want to call this REST API.  I know it's a REST api, but this implies some kind of contract.  I'd rather have a title that reflects the ability to add databases in an automated way and than in the introduction state that it can be done via the rest API.  for now at least until we see more REST api use cases.  This might also change slightly if we go into our 4 editions we spoke about @nishantgeorge

content/ri/using-redisinsight/api.md
nishantgeorge: As we discussed, I've renamed the `dbType` parameter to `connectionType` and also renamed the values:
- `OSS_CLUSTER` => `STANDAONE`
- `OSS_CLUSTER` => `CLUSTER`
- `OSS_SENTINEL` => `SENTINEL`
For enabling vendor and deployment specific features, additional optional parameters may be added in the future.

content/ri/using-redisinsight/api.md
nishantgeorge: Added a new section documenting the endpoint for adding certificates, which returns the ID required to pass here, and liked to that section here.

content/ri/using-redisinsight/api.md
nishantgeorge: Changed the title and added an intro. Works?

nishantgeorge:
@bmansheim @K-Jo These docs should not be released until the next release of RedisInsight where these changes will land.

nishantgeorge:
**PENDING:**
- [ ] GET endpoint for client certificates
- [ ] Something along the lines of a `provider` field with options like `ELASTICACHE`, `REDIS_SOFTWARE`, `REDIS_CLOUD`, etc. (TDB)

PR #694, opened 2020-02-26
Add 5.4.14 release notes

bmansheim:


bmansheim:
@AlonMagrafta Good catch!

PR #697, opened 2020-03-01
Update combine-smaller-strings-to-hash.md

HassanNemir:


bmansheim:
@HassanNemir Thanks for reporting this! It will be updated in the next push to production.

PR #705, opened 2020-03-18
Fix link

imba-tjd:


bmansheim:
Thanks @imba-tjd for catching that!

PR #707, opened 2020-03-19
Fix Hugo update issues

HarunD:
This fixes several errors that appeared after update to new version of Hugo (0.67).
Some warnings about features that will not be supported in the future remain but they are not blocking at the moment, the site can be built. This will be a larger change and will be done in  a separate PR.


bmansheim:
Thanks, Harun. I'll look it over.

bmansheim:
Still got: Page.URL is deprecated and will be removed in a future release. Use .Permalink or .RelPermalink. If what you want is the front matter URL value, use .Params.url

HarunD:
> Still got: Page.URL is deprecated and will be removed in a future release. Use .Permalink or .RelPermalink. If what you want is the front matter URL value, use .Params.url

Yeah, those were the warnings I referred to in PR description:

> Some warnings about features that will not be supported in the future remain but they are not blocking at the moment, the site can be built. This will be a larger change and will be done in a separate PR.

They are not preventing the site from being built and ran at the moment. And the change will be probable be a larger one and more detailed testing will be needed so I suggest to take care of these errors in a separate PR to keep it simpler and decoupled from the changes in this PR.

PR #709, opened 2020-03-23
Update module release notes links

bmansheim:
Redisearch release notes link should be https://docs.redislabs.com/latest/modules/redisearch/release-notes/redisearch-1.4-release-notes/#redisearch-1-4-25  or maybe no need to mention the 1-4-25 but the general link only
The reference to the other modules also doesn't work


bmansheim:
Moved to another branch to fix all instances

bmansheim:
Resolved in: https://github.com/RedisLabs/redislabs-docs/pull/722

PR #710, opened 2020-03-23
Proposing to reference location for socket files in "File Locations"

slemeur:
Socket files are not getting listed in "File Locations" (https://docs.redislabs.com/latest/rs/installing-upgrading/file-locations/) docs and cen be found only in the document about how to change the default location for socket files (https://docs.redislabs.com/latest/rs/installing-upgrading/configuring/change-location-socket-files/)


bmansheim:
Thanks for that @slemeur !
@AlonMagrafta Is there any reason not to list the socket files here?

AlonMagrafta:
@bmansheim I don't see a problem with it.

PR #716, opened 2020-03-30
Update rs-5-4-14-february-2020.md

AlonMagrafta:


AlonMagrafta:
@IAmATeaPot418 
We didn't test it.
I added also build number (20) which wasn't the latest build. And latest build has no problem with this upgrade (directly to 5.6.0). So not worth the effort now.

PR #718, opened 2020-03-30
Add a link to the configuration page

dallastexas92:
It's not immediately obvious that you can customize redisInsight to change things like the default port. If I were a customer, I would look under 'troubleshooting' for how to do this. Therefore, I think a link should be added to the 'configuring RedisInsight' page. I've now had several customers ask if this was possible so I think this would be a great addition!


slemeur:
Thanks that's great improvement for the docs.
Also taking the opportunity to link this PR to https://redislabs.atlassian.net/browse/RI-243

PR #726, opened 2020-04-06
Update docs for upgrade module parameter changes

bmansheim:
@jhershberg Please revew

### DIFF
--- a/content/modules/upgrading-rs.md
+++ b/content/modules/upgrading-rs.md
@@ -49,7 +49,7 @@ This causes a short interruption in the availability of this database across the
 When you upgrade the module for a database, you can either:
 
 - Specify the module arguments to replace the current arguments.
-- Specify the 'keep_module_args' flag to use the current argument.
+- Specify the `keep_args` flag to use the current argument.
 
 1. Connect to the terminal of a node in the cluster
 1. Run `rladmin status` to list the databases on the node.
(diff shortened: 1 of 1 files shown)


content/modules/upgrading-rs.md
```diff
@@ -70,27 +70,31 @@ When you upgrade the module for a database, you can either:
 1. To upgrade the module for the database, run:
 
     ```src
-    rladmin upgrade module db_name <database_name> module_name <module_name> version <new_module_version_number> module_args <module arguments>
+    rladmin upgrade db <database_name> and module module_name <module_name> version <new_module_version_number> module_args "<module arguments>"
     ```
 
+    - If you want to remove the existing module arguments, enter `""` without arguments.
+    - If you want to use the exisintg module arguments, enter `"keep_args"`
+    - If you want to update multiple modules, use the `and module` parameter for each module.
```
AlonMagrafta, line 78: its not only for multiple modules but also for one module you will need "and module" (-i.e. in addition to the db). My comment does necessarily requires changes

content/modules/upgrading-rs.md
```diff
@@ -70,27 +70,31 @@ When you upgrade the module for a database, you can either:
 1. To upgrade the module for the database, run:
 
     ```src
-    rladmin upgrade module db_name <database_name> module_name <module_name> version <new_module_version_number> module_args <module arguments>
+    rladmin upgrade db <database_name> and module module_name <module_name> version <new_module_version_number> module_args "<module arguments>"
     ```
 
+    - If you want to remove the existing module arguments, enter `""` without arguments.
```
jhershberg, line 76: Is this true? Not saying it isn't, just didn't know

content/modules/upgrading-rs.md
```diff
@@ -70,27 +70,31 @@ When you upgrade the module for a database, you can either:
...
+    - If you want to use the exisintg module arguments, enter `"keep_args"`
```
jhershberg, line 77: This could just be me but I find the whole quotation mark thing so weird (it's really not the command line standard) that I would stress "with the quotation marks" or something. But again, this is just my personal take. I'm not a writer.

bmansheim: It's in code style and we have examples so I think it will be OK.

content/modules/upgrading-rs.md
bmansheim: @K-Jo Please confirm.

PR #727, opened 2020-04-06
Warning for nameserver name

bmansheim:
Submitted by @zoharfuchs


AlonMagrafta:
@Paz-Yanover 
I guess this is ok?

Paz-Yanover:
@bmansheim @AlonMagrafta - it seems good to me

Paz-Yanover:
@bmansheim - a message from @zoharfuchs-
"please wait with this. I had a discussion with the customer and did some tests over the weekend.
We do have some issues with the RFC but let's not change our documentation since it isn't exactly like this.
I will update the Jira we have."

PR #732, opened 2020-04-09
make expansion of link more intuitive

vikrammoule-redislabs:
many users do not understand or read the details because it is non-intuitive that the link might be clickable


bmansheim:
@vikrammoule-redislabs Thanks for this feedback. Do you think it would help if the text was red like the other links?

vikrammoule-redislabs:
Maybe, I don't know. I think it is better to write the "Click here" word. Support is experiencing unnecessary repetition of the same instructions.

bmansheim:
@HarunD Please make the expand title red and underline on hover like other links.

bmansheim:
Fixed in #762

PR #735, opened 2020-04-13
Expand on instructions on persistent storage in RS

shahar-biron:
Additional details that explain to customers how to provision and connect external storage volumes for persistence.


bmansheim:
@shahar-biron Please review and comment.

PR #744, opened 2020-04-15
K8s cleanup

alexmilowski:
I went through the current documentation under platforms:

- made the generic documentation for Kubernetes first
- adjusted the titles for the topics under "Getting Started with Redis Enterprise Software on Kubernetes"
- added a description for each topic
- rewrote parts of "Redis Labs Kubernetes Architecture"
- added "Installing the Operator"
- tweaked the title for OpenShift so that OpenShift came first

All the title tweaking was to navigation clear so we don't repeat "Kubernetes" or "Redis Enterprise" more than we need to do so.

"Installing the Operator" defers to the PKS/OpenShift topics for installing the operator on those platforms.


content/platforms/kubernetes/kubernetes-architecture.md
```diff
@@ -1,44 +1,55 @@
...
+Kubernetes is an excellent and widely adopted orchestration platform. While it can provide the building blocks for a variety of applications, it was not designed to deal with all the nuances associated with operating Redis Enterprise. Therefore, without help, the Kubernetes cluster cannot react accurately to internal Redis Enterprise events, failure conditions, or user requests.
 
-To overcome these issues, Redis Labs created a layered architecture approach that splits responsibilities between operations Kubernetes does well, procedures Redis Enterprise Cluster excels at, and the processes both can orchestrate together. The figure below illustrated this layered orchestration architecture:
+The orchestration within a Kubernetes cluster runs outside the Redis Cluster deployment. As such, it is not aware of internal events and so can not respond to trigger failover events. For example, the Kubernetes cluster will not be aware of split network scenarios that affect Redis Enterprise clusters.
```
amiramm, line 20: Instead of addressing split-network scenario - K8s is responsible for the network and we can't necessarily address split network from within the cluster or Operator - so I would stick with the failover example.

content/platforms/kubernetes/kubernetes-architecture.md
```diff
@@ -1,44 +1,55 @@
...
+Redis Enterprise is not only great as an in-memory database but also extremely efficient in the way it uses persistent storage, even when the user chooses to configure Redis Enterprise to write every change to the disk. Compared to a disk-based database that requires multiple interactions (in most cases) with a storage device for every read or write operation, Redis Enterprise uses a single IOPS, in most cases, for a write operation and zero IOPS for a read operation. As a result, the performance of Redis Enterprise is not impacted significantly by using network-based storage. Moreover, significant performance improvements over other similarly deployment databases are typical in Kubernetes environments, as illustrated in the figures below:
```
amiramm, line 45: @alexmilowski `the performance of Redis Enterprise is not impacted significantly by using network-based storage`
We require block persistent storage be attached. This implies we can provide the same performance with a SAN or NAS, which isn't necessarily the case.

content/platforms/kubernetes/kubernetes-architecture.md
```diff
@@ -1,44 +1,55 @@
...
+Redis Labs took a different approach to managing Redis Enterprise over the Kubernetes environment. Each Pod includes multiple Redis Enterprise instances (multiple services). By deploying doing so, this allows us to better utilize the hardware resources used by the Pod such as CPU, memory, and network while keeping the same level of isolation. This is shown on the left side of the figure below.
+
+In our experience, we found that the traditional method of deploying a Redis Enterprise database in Kubernetes as separate pods, preserving a dedicated amount of computed resources (CPU & memory), that contain a single Redis Enterprise instance was notably inefficient. Redis Enterprise is exceptionally fast and in many cases can use just a fraction of the CPU resources to deliver the requested throughput. Furthermore, when running a Redis Enterprise Cluster with multiple Redis Enterprise instances across multiple Pods, the Kubernetes network can quickly become the deployment’s bottleneck. This is show on the right side of the figure below.
```
amiramm, line 53: @alexmilowski I'm not clear on why you switched these around. It reads much better when we end this section on the positive note of what RE can do rather than the negative.

content/platforms/kubernetes/kubernetes-installing.md
```diff
@@ -0,0 +1,150 @@
...
+[GitHub](https://github.com/RedisLabs/redis-enterprise-k8s-docs) and the
+implementation is published as a Docker container. The operator
+definitions are packaged as a single generic YAML file called
+[bundle.yaml](https://github.com/RedisLabs/redis-enterprise-k8s-docs/blob/master/bundle.yaml).
+
+The raw bundle can be downloaded via `curl`:
+
+```
+
+curl -O https://raw.githubusercontent.com/RedisLabs/redis-enterprise-k8s-docs/v5.4.14-7/bundle.yaml
```
amiramm, line 54: @alexmilowski we shouldn't point to a specific version, otherwise, we'll have to update this doc every 6-8 weeks as we release new builds and versions.

content/platforms/kubernetes/kubernetes-persistent-volumes.md
```diff
@@ -1,6 +1,7 @@
 ---
-Title: Deploying Kubernetes with Persistent Volumes in Operator-based Architecture
-description: 
+Title: Deploying with Persistent Volumes
+description: A Redis Enterprise Cluster deployed with the operator should include
```
amiramm, line 3: @alexmilowski this is somewhat misleading since we implicitly do that even if it isn't specified in the spec. On a different note, I would stick with a description of the article rather than providing directives in the `Description` field.

content/platforms/kubernetes/kubernetes-architecture.md
alexmilowski: Split networks was in the original content. I can remove it.

content/platforms/kubernetes/kubernetes-architecture.md
```diff
@@ -1,44 +1,55 @@
...
 
-Operator allows Redis Labs to maintain a unified deployment solution across all Kubernetes environments, i.e., RedHat OpenShift, Pivotal Container Services (PKS), Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), Amazon Elastic Container Service for Kubernetes (EKS) and vanilla Kubernetes. Statefulset and anti-affinity guarantee that each Redis Enterprise node resides on a Pod that is hosted on a different VM or physical server. See this setup shown in the figure below:
+Operator allows Redis Labs to maintain a unified deployment solution across all Kubernetes environments, i.e., RedHat OpenShift, Pivotal Container Services (PKS), Google Kubernetes Engine (GKE), Azure Kubernetes Service (AKS), Amazon Elastic Container Service for Kubernetes (EKS) and vanilla Kubernetes. By utilizing various Kubernetes features like [StatefulSets](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) and anti-affinity rules, the operator guarantees that each Redis Enterprise node resides on a Pod that is hosted on a different VM or physical server. See this setup shown in the figure below:
```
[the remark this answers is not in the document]
alexmilowski: Let's go with OSS

content/platforms/kubernetes/kubernetes-architecture.md
alexmilowski: "As a result, the performance of Redis Enterprise is not impacted in the same way more I/O intensive databases are by using network-based storage."

That is the point, right?

content/platforms/kubernetes/kubernetes-architecture.md
alexmilowski: I think we need to introduce what we did and the benefits without dwelling on the "wrong" approaches. I think I can merge these two.

content/platforms/kubernetes/kubernetes-installing.md
alexmilowski: We also can't have them pull from the master because that could be a work in progress that is not compatible with the desired version. Maintaining the correct current version might be something we need to handle ... or we change the URL to have {version} in there and give a procedure for finding the version (i.e., looking at the version tags in GitHub or syntax "v" + version)

content/platforms/kubernetes/kubernetes-persistent-volumes.md
alexmilowski: "Configuring Persistent Volumes" ?

The descriptions are autogenerated by taking the first few sentences from the markdown. It can generate nonsense quite easily. Summaries need to be written and not generated.

amiramm: How are descriptions auto-generated? is that a new feature? can we disable it? I thought you wrote in your PR that you "added descriptions"?

content/platforms/kubernetes/kubernetes-installing.md
amiramm: having a placeholder is the better option here I believe. Versioning of docs as a whole would be the preferred solution here. @banker or @bmansheim can comment on when that is planned.

content/platforms/kubernetes/kubernetes-architecture.md
amiramm: @alexmilowski - agree we should avoid calling is "wrong". Need to compare and contrast the "OSS redis server as a Pod" approach with "Redis Enterprise Cluster node as a Pod" approach

content/platforms/kubernetes/kubernetes-architecture.md
amiramm: "by using Kubernetes persistent volumes" (and implicitly, those should be block storage devices).

content/platforms/kubernetes/kubernetes-installing.md
alexmilowski: I came up with a different solution based on using the GitHub API and releases. You can now pull the latest release with the script or look up a specific release tag yourself.

content/platforms/kubernetes/kubernetes-persistent-volumes.md
alexmilowski: They just pull some text from the targeted file ... the first few sentences excluding any section titles.

alexmilowski:
I'm not sure what you mean by "generic document". All my changes are under "Kubernetes" and I left the content in "_index.md" all commented out. The only thing I changed to make first was that "Getting Started with Redis Enterprise Software on Kubernetes" is first in the navigation bar on the left.

alexmilowski:
I have made a variety of changes to address the comments.

bmansheim:
@amiramm Is the ball in your court?

bmansheim:
@amiramm @alexmilowski Where do we go with this one?

PR #745, opened 2020-04-15
Update _index.md

kerneljake:
Fix alpha versioning


kerneljake:
The versioning was updated in RED-30061, and there is a [link](https://raw.githubusercontent.com/RedisLabs/redis-enterprise-k8s-docs/5.4.10-8/crds/app_v1_redisenterprisecluster_cr.yaml) preceding this block to an example that says "v1", so am I assuming this block needs to be updated.

PR #748, opened 2020-04-16
K8s OpenShift Reorg + OperatorHub

alexmilowski:
I broke the various tasks into different sections and preserved the existing content or updated as necessary. There are now new sections on installing via CLI, installing via OperatorHub, Creating Clusters and Databases, and the existing Active-Active/CRDB content.


alexmilowski:
This has been superseded by Pull request #760

PR #751, opened 2020-04-18
Emphasize replication for active-active

kashish-redislabs:
After discussion with Zohar (Director, Global Support) we'd like to update the verbiage on Replication to encourage customers to enable it in the majority of the cases. For any queries on this change, please reach out to Zohar.


bmansheim:
@AlonMagrafta @zoharfuchs Should we consider making replication enabled by default for active-active databases?

AlonMagrafta:
I think we can consider it. @bmansheim you mean in practice when creating new CRDB right? Not just Docs.
@Paz-Yanover FYI

content/rs/administering/database-operations/create-active-active.md
```diff
@@ -68,8 +68,7 @@ Every instance of a CRDB can receive write operations, and all operations are [s
 
     {{% /note %}}
 
-    - **Replication** - We recommend that you use intra-cluster replication to create slave shards in each CRDB instance.
-        The intercluster synchronization is most efficient when it reads from slave shards.
+    - **Replication** - Please enable intra-cluster replication to create slave shards in each CRDB instance. The intercluster synchronization is most efficient when it reads from slave shards and also supports Slave HA for redundancy. CRDB without slave shards isn’t recommended.
```
Paz-Yanover, line 71: Let's link to the slave HA documentation

bmansheim:
@AlonMagrafta Correct. If we turn it on by default then we don't need the docs to "encourage" customers to turn it on. In the meantime I'll have to update the docs with the encouragement.

cc: @Paz-Yanover

PR #759, opened 2020-04-21
featureset version in RN5-6-0

AlonMagrafta:


bmansheim:
@AlonMagrafta Seems to me that we want everyone who upgrades to 5.6 with A-A dbs to update the feature set. If so, I think we can simplify the Upgrade statement and that we don't need the statement in the HLL section.

AlonMagrafta:
@bmansheim @Paz-Yanover and I think it's important to mention it this time twice - as a general note presenting "featureset version" and specifically for the HLL on AA feature.

PR #764, opened 2020-04-26
Non-existing page for persistence

tom-andrews:
The [persistence](https://redislabs.com/redis-features/persistence) URL (in multiple places) leads to an non-existing page.
Not sure exactly what that page was, but I believe the most suitable new URL will be - https://redislabs.com/redis-enterprise/technology/durable-redis-2/


bmansheim:
I think it would be better to move the information from the marketing site to the docs. What do you think, @banker ?

banker:
@bmansheim I think we need both. This information definitely needs to be on the doc site, but it needs to be in the form of documentation. Whether it still exists as a marketing page is up to marketing. Created a ticket here: https://redislabs.atlassian.net/browse/DOC-206

PR #765, opened 2020-04-26
fix reference in section Redis Client Support

shahar-biron:
There is no client related details in the HW requirements page - it is written in supported-clients-broswers page instead


bmansheim:
@shahar-biron Thanks for the PR. Can you confirm that my changes are OK?

shahar-biron:
Thanks @bmansheim looks ok, the main point is making the proper reference. Thanks.

bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/753
Discovery clients are now listed on the page.

PR #766, opened 2020-04-27
RED-35577 Demote master node during maintenance

bmansheim:


content/rs/administering/cluster-operations/maintenance-mode.md
```diff
@@ -88,6 +86,18 @@ To turn maintenance mode on and prevent slave shard migration, on one of the nod
 rladmin node <node_id> maintenance_mode on keep_slave_shards
 ```
 
+### Demoting a master node in maintenance mode
+
+When you turn maintenance mode on for a master node in order to do maintenance that interrupts connectivitiy to the master node,
+you can demote the master node so that another node becomes the master.
+When you demote the master node, the connectivity outage caused is shorter than the outage caused when the cluster detects that the master not available.
```
AlonMagrafta, line 93: Do we want to say something more positive? Like:
"When you demote the master node, re-election and its availability are higher than letting the cluster to detect that the master not available."
Not the best sentence but you probably get the idea

PR #774, opened 2020-04-29
RedisInsight - Add documentation for optional ACL username field in add instance form

himanikhurana10:


slemeur:
@himanikhurana10 Could you also update the section under TLS ? Item #4 would need the screenshot to be updated - and also explanations about how to proceed in the situation of TLS + ACL

PR #749, opened 2020-04-17
Update upgrading-kubernetes-with-operator.md

barry-waldbaum-rl:
kubectl rollout status sts

doesn't work with my client:

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-
03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-gke.27", GitCommit:"145f9e21a4515947d6fb10819e5a336aff1b6959", GitTreeState:"clean", BuildDa
te:"2020-02-21T18:01:40Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}

PR #780, opened 2020-04-30
Remove Services Rigger from Kubernetes upgrade

laurentdroin:
Unless I misunderstand something, the services rigger is upgraded with the REC, not when applying the bundle.

### DIFF
--- a/content/platforms/kubernetes/upgrading-kubernetes-with-operator.md
+++ b/content/platforms/kubernetes/upgrading-kubernetes-with-operator.md
@@ -62,16 +62,16 @@ These diagrams show how a rolling update works:
 - Each hexagon represents a node
 - Each box represents a pod
 
-![kubernetes-rolling-updates]( /images/rs/kubernetes-rolling-updates.png )
+![kubernetes-rolling-updates](/images/rs/kubernetes-rolling-updates.png?width=500)
 
 The pods are updated one by one in the diagram, starting from left to right.
 Each pod is updated after the last one completes successfully.
 
-![kubernetes-rolling-updates-newapp]( /images/rs/kubernetes-rolling-updates-newapp.png )
+![kubernetes-rolling-updates-newapp](/images/rs/kubernetes-rolling-updates-newapp.png?width=500)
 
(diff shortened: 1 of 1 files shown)


content/platforms/kubernetes/upgrading-kubernetes-with-operator.md
```diff
@@ -75,4 +75,4 @@ Each pod is updated after the last one completes successfully.
 
 The pods in the StatefulSet are updated in reverse ordinal order.
 The Kubernetes controller terminates each pod and waits for it to transition to `Running` and then to `Ready`.
-The it updates the next pod until all pods are updated and the upgrade process is completed.
\ No newline at end of file
+The it updates the next pod until all pods are updated and the upgrade process is completed.
```
bmansheim, line 78: What's "The it"?

laurentdroin: Hmmm... I don't remember adding that line. I just wanted to remove the reference to the Services Rigger (line 38). Let me think about it some more.

laurentdroin: This is strange. I don't think that line is from me. This is not my style. I don't think I would use the expression "update a pod". Still checking.

laurentdroin: The change was already there in PR #749

alexmilowski: I think this is out of date now so the PR needs to be updated. We have confirmed this is the correct behaviour and so removing "Service Rigger" from the step is technically correct.

alexmilowski: Must be a line-ending change on that last line. I think we should change that line to:

"Once a pod is updated, the next pod is terminated and updated. Once all the pods are updated, the upgrade process is completed."

bmansheim: Done

PR #789, opened 2020-05-03
6.0 life cycle

AlonMagrafta:
@Paz-Yanover WDYT - April to Oct or should we do May to Nov? I think we can do April since it was our official release


bmansheim:
@AlonMagrafta Waiting for final approval

AlonMagrafta:
Updated to May-Nov.
@bmansheim approved now

PR #791, opened 2020-05-05
DOC-204 Developing with Active-Active streams

bmansheim:


content/rs/developing/crdbs/developing-streams-crdb.md
```diff
@@ -115,158 +123,155 @@ Using XREAD to iterate through stream entries is a popular pattern (Very similar
...
+| t4   | --- sync ---                                      | --- sync ---             |
+| t5   | XRANGE x - + --> [110-0]                          | XRANGE x - + --> [110-0] |
+| t6   | XINFO GROUPS x -> [g1]                            | XINFO GROUPS x -> [g1]   |
+| t7   | XINFO CONSUMERS x -> [Alice]                      | XINFO CONSUMERS x -> []  |
+| t8   | XPENDING x g1 - + --> [110-0]                     | XPENDING x g1 - + --> [] |
+
+As you can see, Location 2 seems to be completely unaware of any group-related activity and it seems that redirecting the XREADGROUP traffic from Location 1 to Location 2 will result in reading the same entries that were already read.
 Given that the usage pattern is standard (Every entry that was read using XREADGROUP is acknowledged with XACK after processing it) we have a way to tackle this issue.
 
 Note that using XREADGROUP across instances can result in instances reading the same entries because CRDT Streams is designed to handle “at least once” reads (or have only one reader at a time).
```
bmansheim, line 225: @guybe7 What did you mean by "we have a way to tackle this issue"?

guybe7: the issue is the fact the Location 2 is unaware of any group-related activity in Location 1. it poses this problem: "redirecting the XREADGROUP traffic from Location 1 to Location 2 will result in reading the same entries that were already read"
so the issue leads to a problem and the solution is described in the section "### XREADGROUP redirection"

(i thin that the paragraph "	Note that using XREADGROUP across instances.." was not there originally so the transition from "here is a problem" to "the section that describes the solution" was smoother)

PR #793, opened 2020-05-06
Add upgrade note for 6.0

AlonMagrafta:


bmansheim:
@AlonMagrafta Should we remove the note for 5.2? It went out of support in Dec 2019.

AlonMagrafta:
I don't think so, I thinks it's too early. We even have customers with 5.0.2 (saw several related tickets about it). So let's wait a bit. Thanks for pointing out.
@bmansheim

PR #795, opened 2020-05-06
Slave HA prerequisites

bmansheim:


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/796

PR #798, opened 2020-05-07
DOC-227 Add commands to change inactivity duration

bmansheim:


IAmATeaPot418:
I feel like the rest API documentation portion of it is in the wrong place, but regardless of that,  it doesn't show the authentication structure of the command, so anyone trying to copy-paste it wouldn't be able to successfully authenticate.

PR #799, opened 2020-05-07
Import page with s3-compatible

AlonMagrafta:
@bmansheim - maybe we need this:
{{< embed-md "backup-locations.md" >}}
?


bmansheim:
@AlonMagrafta The text is geared toward backup, not import. We'll just add the S3 note.

PR #801, opened 2020-05-07
LDAP update

IAmATeaPot418:


banker:
@IAmATeaPot418 Can you check my edits?

IAmATeaPot418:
@AlonMagrafta - that's already in there in the note.

{{% note %}}
LDAP authentication is not yet supported for Redis ACL Users. LDAP authentication is only supported for the control plane.
{{% /note %}}

IAmATeaPot418:
@banker you're improvements make it much clearer. There are two things I'd change with them:

1. `MECHANISM`S should have the apostrophe after the s. Its the exact name of the variable, not the possessive.
2.  (or to recycle `saslauthd`) - the step outlined is to recycle saslauthd for the changes to take effect, I'm not sure I understand what the parenthesis mean. Do you want to modify directly or should I?

ushachar:
@bmansheim / @IAmATeaPot418  / @banker / @AlonMagrafta  I'm not sure why this edit was made. It seems to point to a completely incorrect method.
- There shouldn't be any need to manually edit *any* file. If there is, it's a bug.
- Please do not edit /etc/ files. If you want to use the existing file as a template, copy it aside and edit that.

bmansheim:
@AlonMagrafta I can change the instructions to save a copy of the original file but do you want to open a task for R&D to rework this process so it doesn't require manually editing a file in etc?

PR #802, opened 2020-05-08
ACL updates

IAmATeaPot418:
@banker @AlonMagrafta @bmansheim - I think this will help simplify the ACL documentation for our users and make it a bit more workflow-based and explanatory.

### DIFF
--- a/content/rs/administering/access-control/user-roles.md
+++ b/content/rs/administering/access-control/user-roles.md
@@ -59,6 +59,9 @@ Additional multi-keys commands restrictions such as when using OSS Cluster mode 
 Redis ACLs are defined by a [Redis syntax](https://redis.io/topics/acl#acl-rules) where you specify the commands or command categories that are allowed for specific keys. Redis Enterprise only allows the configuration of commands, command categories and allowed keys. 
 
 Redis Modules do not have a command category, but may be configured using their standard command name with Redis ACL syntax.
+{{% note %}}
+Module's internal processes are not retricted by the Redis ACLs to allow their proper functioning.
+{{% /note %}}
 
 **In Redis Enterprise, you can:**
 
--- a/content/rs/administering/access-control/user-roles.md
+++ b/content/rs/administering/access-control/user-roles.md
@@ -55,8 +55,8 @@ Redis Modules do not have a command category, but may be configured using their 
 
 **In Redis Enterprise, you can:**
 
-- **Include commands and categorie**s with the + prefix for commands or +@ prefix for a categories
-- **Exclude commands and categories** with the - prefix for commands or -@ prefix for a categories
+- **Include commands and categorie**s with the + prefix for commands or +@ prefix for a category
+- **Exclude commands and categories** with the - prefix for commands or -@ prefix for a category
 - **Include keys or key patterns** with the ~ prefix
 
 **To define database access control, you can either:**
--- a/content/rs/administering/access-control/user-roles.md
+++ b/content/rs/administering/access-control/user-roles.md
@@ -18,11 +18,11 @@ Role-Based Access Control allows you to scale your Redis deployments while minim
 
 ### Configuring Roles & Users
 
-In **access control > roles**, you can configure user roles with:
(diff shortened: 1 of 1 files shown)


content/rs/administering/access-control/user-roles.md
```diff
@@ -5,83 +5,73 @@ weight: $weight
...
+To create a user role that only has access to the database, assign the user the management role **None**
+{{% /note %}}
 
-To control user access to Redis database commands and keys,
-you must define Redis ACLs that specify the commands that users can run and keys that the commands can apply to.
-Then, in the user role, select the databases that the users can access and Redis ACL that controls user access to those databases.
+### Database Access Controls
 
-A-A cannot assign ACLs on creation
+ACLs allow you to control what level of access each user has in Redis. With ACLs, you can specify which commands specific users can execute and which keys they can access. This allows for much better security practices: you can restrict any given user’s access to the least level of privilege needed.
```
banker, line 39: @IAmATeaPot418 This seems to imply that you set access on a per-user basis. I think this needs some rewording.

AlonMagrafta: How about:
**Redis** ACLs allow you to control what level of access each **role and** user have in Redis. With **Redis** ACLs, you can specify which commands specific **roles and** users can execute and which keys they can access. This allows for much better security practices: you can restrict any given user’s access to the least level of privilege needed.

content/rs/administering/access-control/user-roles.md
```diff
@@ -5,83 +5,73 @@ weight: $weight
...
-    In Redis:
-    - These ACL subcommands are blocked: LOAD, SAVE, SETUSER, DELUSER, GENPASS, LOG
-    - These ACL subcommands are allowed: LIST, USER, GETUSER, CAT, WHOAMI, HELP
-- The MULTI, EXEC, DISCARD commands are always allowed, but ACLs are enforced on MULTI subcommands.
-- External users cannot authenticate with databases.
-- When you run multi-key commands on multi-slot keys, the return value is `failure` but the command runs on the keys that are allowed.
-
+1. The ACL subcommands LOAD, SAVE, SETUSER, DELUSER, GENPASS and LOG are blocked
+2. The MULTI, EXEC, DISCARD commands are always allowed, but ACLs are enforced within a transaction
+3. When you run multi-key commands on multi-slot keys, the return value is `failure` but the command runs on the keys that are allowed.
```
AlonMagrafta, line 46: Let's add here that you can edit this as configuration per DB to block or allow such cases.
Using rladmin:
rladmin tune db <db:id> mkms <enabled/disabled>
Also important: inherent commands restrictions for multi-keys commands such as in OSS Cluster mode or in Active-Active will apply even when this setting (mkms) is turned ON.

IAmATeaPot418: @AlonMagrafta - would you be able to take a swing at this. I don't understand it fully enough to do it the right level of justice.

content/rs/administering/access-control/user-roles.md
IAmATeaPot418: I like Alon's wording. Will add a commit with that if you agree Kyle.

banker: Fine by me @IAmATeaPot418

IAmATeaPot418:
@AlonMagrafta - LMK what you think.

content/rs/administering/access-control/user-roles.md
AlonMagrafta: I pushed the new wording in the last commit

content/rs/administering/access-control/user-roles.md
AlonMagrafta: Took it :)

AlonMagrafta:
@IAmATeaPot418 @banker 
I pushed two more commits

IAmATeaPot418:
@banker @bmansheim - thoughts?

bmansheim:
I'll assume these changes will be included in the RS and RC Security overhaul.

PR #805, opened 2020-05-11
Style and format for OpenShift

bmansheim:


bmansheim:
@alexmilowski I recommend we use this to resolve the conflict:

The OperatorHub is a feature of OpenShift 4.x that enables browsing a catalog
of open-source and vendor-provided operators.

You can install the operator using the Command-Line Interface (CLI), but the
OperatorHub is an easy way to help you to manage the operator and apply
upgrades. It also provides an easy way to create clusters.

The operator installation process is:

1. Install the operator using the OperatorHub.
1. Create clusters with a Custom Resource Definition (CRD) from the Operator view
   in the OperatorHub.
1. Create databases in Redis Enterprise.

Now the databases are exposed as Kubernetes services accessible by other
application workloads in your K8s cluster.

bmansheim:
@alexmilowski Can you review this?

PR #809, opened 2020-05-12
Update developing-streams-crdb.md

Paz-Yanover:


banker:
@Paz-Yanover I've fixed all of this in https://github.com/RedisLabs/redislabs-docs/pull/813

PR #819, opened 2020-05-15
Fix note styling

banker:
Notes don't need a heading or italics. They're already sufficiently called out. This is more readable.


bmansheim:
@banker Good idea. Should we get rid of the heading for the other notices (https://docs.redislabs.com/staging/fix-note-styling/cheatsheet/#informative-notices) also? We currently only use note and warning, but we also have info and tip available.

bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/commit/96995c6e4d83c9a9eb3d7d3eabdf967f944620e0

banker:
@bmansheim This is not entirely fixed in the referenced issue

PR #834, opened 2020-05-20
fixed links

barry-waldbaum-rl:


alexmilowski:
@bmansheim This fixes the links.

PR #846, opened 2020-05-27
Fix hugo update issues

bmansheim:


bmansheim:
Reverted. Resubmitted in #847

PR #860, opened 2020-06-04
Only works for 5.6+

kurtfm:
If you don't have version 5.6 you won't be able to find rl_uninstall.sh.  There are a lot of people still on previous versions that are upgrading or administering and need to know how to uninstall using the older method.

### DIFF
--- a/content/rs/installing-upgrading/uninstalling.md
+++ b/content/rs/installing-upgrading/uninstalling.md
@@ -15,3 +15,27 @@ To uninstall RS from a node, run:
 ```src
 rl_uninstall.sh
 ```
+
+For RS versions below 5.6 you will need to utilize the following methods:
+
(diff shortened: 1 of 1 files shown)


bmansheim:
@kurtfm Hey Kurt. I'm glad you brought this up. The principle is that the latest docs are for version 6.0 so that if someone wants to uninstall a previous version they need to go to the docs for that version to get the procedure. Do you think customers will not understand this and will create support tickets because they can't find uninstall instructions for their version in the latest docs?

banker:
We can always add a note that this changed recently and then link to the older instructions.

kurtfm:
@bmansheim and @banker I did not even notice that we now have versioned docs, apologies.  Is it at all possible to show the version at the top of the specific doc and not just the menu?  I don't know for sure, but since the SA asking me and myself didn't see this my guess is that customers may miss it as well.

PR #865, opened 2020-06-09
Add explanation how to retrieve CRDB-GUID

shahar-biron:
Needed for novice Active/Active admins.


bmansheim:
@banker There are rendering issues in this PR. They'll be handled in another PR.

PR #872, opened 2020-06-14
Update AWS backup screenshot and encryption requirements

yorozen:
Need to update AWS screenshots to current AWS UI, Need to add that AWS-KMS Encrypting is not supported.
And only  None or AES-256 are supported?


bmansheim:
Hi @yorozen . Good point. Please make the text changes and add the screenshot to the static/images.

PR #883, opened 2020-06-17
Updated microsoft blobs

yorozen:
Used links to Microsoft in order to reduce mistakes


bmansheim:
@yorozen Good idea to reduce the amount of upkeep with the 3rd party vendor docs. I'll think about how we can make this more efficient and accurate.

kaitlynmichaelredislabs (approved):

PR #884, opened 2020-06-17
DOC-282 Change microseconds to seconds for latency metrics

bmansheim:


bmansheim:
Approved by Mike Clapper

PR #885, opened 2020-06-18
RedisInsight strong admin password

bmansheim:


rromanchuk:
👍 thanks, that's what i figured it met. Finally just created a tunnel in, something that has been on my todo list for a verrrrrry long time. Ps: this is a nice product, i've been waiting for something like this for what feels like over a decade.

bmansheim:
Thanks for the nice comment. Glad you got it working!
cc: @slemeur

PR #895, opened 2020-06-23
Prevent OS from using Redis ports

yanivhershkovich:
In Pre-Requisites, added a new bullet with a recommendation for setting the OS local port usage as we had few tickets where Redis services couldn't start due to port conflict.
As a result, customers asked why this is not showing in our documentation.
Rlcheck which runs post-installation also checks for the configured port range but it doesn't do more than a warning that can easily be overlooked.

I thought this page would be the best place but if there's a better place, we can move it there.
Please verify that I added it as a bullet and didn't mess up the page's formatting.


bmansheim:
@yanivhershkovich Looks pretty good. I changed the wording some. Is it still OK?

yanivhershkovich:
@bmansheim yes, looks good. Thanks!

PR #912, opened 2020-07-06
DOC-308 Distributed syncer

bmansheim:


AlonMagrafta:
@bmansheim This PR doesn't include any information about the Distributed Syncer

bmansheim:
@AlonMagrafta @banker It doesn't seem like /rs/administering/designing-production/networking/multiple-active-proxy.md belongs in networking. Where do you think it should go?
I'm thinking that it and the new distributed syncer article should go in /rs/administering/cluster-operations/.

content/rs/administering/cluster-operations/synchronization-mode.md
```diff
@@ -0,0 +1,50 @@
+---
```
bmansheim, line 1: @AlonMagrafta Please review this article

content/rs/administering/active-passive.md
```diff
@@ -177,3 +183,9 @@ which shard/s the command refers.
 
 The source and destination can have different shard counts and functions
 for placement of keys.
+
+### Syncer in Active-Passive Replication (Replica Of)
```
AlonMagrafta, line 187: I don't understand this paragraph. We need to discuss it

content/rs/administering/cluster-operations/synchronization-mode.md
```diff
@@ -0,0 +1,50 @@
...
+
+    The output of the status command shows the list of endpoints on the cluster and the proxy policy for the endpoint.
+
+    ```sh
+    ENDPOINTS:
+    DB:ID       NAME      ID                        NODE          ROLE                                SSL
+    db:1        db        endpoint:1:1              node:1        all-master-shards                   No
+    ```
+
+    If the proxy policy is `single`, configure the policy to `all-nodes` or `all-master-shards` according to your needs with the command:
```
AlonMagrafta, line 34: Let phrase it
"policy (role field)"

AlonMagrafta: Or 
"policy (role)"

content/rs/administering/active-passive.md
bmansheim: What are your thoughts?

AlonMagrafta: Flush before full sync
When replication backlog is full.

In DR - before getting to work only from the replicaOf - disable the "replicaOf" / replication so we won't get flushed

bmansheim: I addresses this in commit c08158ea1d702f028c307bfae0e3b3233e14bc22. What do you think?

content/rs/administering/active-passive.md
```diff
@@ -179,3 +185,21 @@ which shard/s the command refers.
...
+the destination database flushes all of the data from its memory and starts synchronizing the data again.
+This process is called a **full sync**.
+
+For example, if the database updates for the destination databases
+that are stored by the destination database in a synchronization backlog exceed their allocated memory,
+the source database starts a full sync.
+
+{{% warning %}}
+To avoid a full sync in a disaster recovery scenario,
+make sure that you disable **Replica Of** before you direct clients to the destination database.
```
AlonMagrafta, line 204: We need to emphasize here that there is no problem reading from replica-of
It's the writes which are dangerous (for potential data loss)

PR #940, opened 2020-07-22
DOC-302 Add upgrade latest_with_module

bmansheim:


content/modules/add-module-to-cluster.md
```diff
@@ -70,34 +70,58 @@ To upgrade a module that is installed on a cluster:
...
+    - `rladmin status modules` - Shows the latest modules available on the cluster
+        and the modules used by databases.
+    - `rladmin status modules all` - Shows all of the modules available on the cluster
+        and the modules used by databases.
+
+1. To upgrade modules for the database, run:
 
     ```sh
-    rladmin upgrade db <database_name> and module module_name <module_name> version <new_module_version_number> module_args "<module arguments>"
+    rladmin upgrade db <database_name | database_ID> and module module_name <module_name> version <new_module_version_number> module_args "<module arguments>"
```
AlonMagrafta, line 83: I know database_id works. Has database_name been verified with R&D?

content/modules/add-module-to-cluster.md
```diff
@@ -70,34 +70,58 @@ To upgrade a module that is installed on a cluster:
...
     ```
 
-    - If you want to remove the existing module arguments, enter `""` without arguments.
-    - If you want to use the existing module arguments, enter `"keep_args"`
+    - To upgrade the database to use the latest version of Redis and the latest version of the modules used by the database without changing the module arguments,
+        use the `latest_with_modules` parameter.
```
AlonMagrafta, line 87: I think since this will be a primary way to upgrade, we need to add here the command explicitly. Add something like:
"
Run:
rladmin upgrade db <db:id> latest_with_modules
"

content/modules/add-module-to-cluster.md
bmansheim: Verified with Guy and on live system

PR #951, opened 2020-07-23
removed extra bracket that broke link

amiramm:


amiramm:
@bmansheim shouldn't this type of issue break CI?

bmansheim:
@amiramm We haven't been as focused on tools as I would like, but there is hope.

PR #952, opened 2020-07-24
Added a script to install the dependencies

viragtripathi:
Added a simple script to install the dependencies to avoid the manual steps which can lead to errors.

### DIFF
--- a/content/modules/redisgears/installing-redisgears.md
+++ b/content/modules/redisgears/installing-redisgears.md
@@ -20,20 +20,23 @@ On each node in the Redis Enterprise cluster:
 
 1. [Download](https://redislabs.com/download-center/modules/) the RedisGears Module - Dependencies Package from the Redis Enterprise Software section of the Downloads page.
 1. Copy the dependencies package to a node in your cluster.
-1. Run these commands as root:
 
-    ```sh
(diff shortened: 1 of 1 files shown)


bmansheim:
@K-Jo We can close this because in 6.0.12 there is no need to install dependencies, correct?

bmansheim:
@K-Jo Looks like there will be dependencies until 6.0.16. Should we merge this until then or just wait?

content/modules/redisgears/installing-redisgears.md
```diff
@@ -20,20 +20,23 @@ On each node in the Redis Enterprise cluster:
...
-    # chown -R $osuser $modulesdatadir/rg
-    ```
-
-{{< note >}}
-- `<version>` - The version number in the format `x.y.z`.
-- `<version_integer>` - The version number in integer format `xxyyzz` or you can calculate it as `10000*x + 100*y + z`.
-- You must also run these commands on new nodes before you join them to an existing cluster.
-{{< /note >}}
+1. Save and Run this script as root:
+cat install_rg_dependencies.sh
```
K-Jo, line 25: This should be a code snippet

K-Jo (changes requested):

bmansheim:
This behavior will change in 6.0.12.

PR #956, opened 2020-07-29
Update product-lifecycle.md

martinez099:
EOL of RSv5.4 was last end of year.


AlonMagrafta:
@martinez099 @bmansheim 
Each version is supported for 18 months, meaning 5.4 was due to EOL end of May 2020, but we extended it because 5.6 was released later than expected.
Therefore until EOY 2020.

PR #962, opened 2020-08-04
Fix typo for importing data from Redis

yanivhershkovich:
In the import from a Redis server, fixed a typo indicating that a user should put RDB details to source DB details [see edited line].

Another point is that the section mentions "import a dataset from any publicly available Redis server" but this is not accurate.
As far as I know, import from a Redis Labs DB is not allowed. I'm not sure if there are other restrictions.
Need to check with PM to clarify that point.
I tried importing from both RS and Essentials (both publicly available) and I couldn't.

### DIFF
--- a/content/rc/how-to/importing-data.md
+++ b/content/rc/how-to/importing-data.md
@@ -25,7 +25,7 @@ To import a dataset from any publicly available Redis server:
 
 1. In the Redis Cloud management console, go to the database that you want to import into.
 1. Click ![Import](/images/rc/icon_import.png#no-click "Import").
-1. Enter the details for the RDB file:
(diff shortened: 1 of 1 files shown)


AlonMagrafta:
@bmansheim It's cloud docs, I guess @joniredis needs to review this.
I can only guess that the "import a dataset from any publicly available Redis server" might be referring to OSS, but I don't know it for sure.

joniredis (approved):

bmansheim:
@joniredis Can you relate to Yaniv's comment about importing from RS and Essentials?

bmansheim:
@joniredis Please check if importing can be done from publicly available RL.

PR #964, opened 2020-08-04
Add dropdown for Platforms

bmansheim:
This dropdown is so that customers can select between the 6.x version of the docs which is the latest and the 5.x version of the docs that is in /5.6.

I tried just re-using the RS dropdown but it didn't work for me.


bmansheim:
Also need to update the bucket policy to redirect 5.2/platforms and 5.4/platforms to 5.6/platforms.

PR #966, opened 2020-08-05
Spelling and grammar check

bmansheim:


bmansheim:
I used LanguageTool (http://wiki.languagetool.org/http-server and https://marketplace.visualstudio.com/items?itemName=davidlday.languagetool-linter) to find some more spelling mistakes.
I installed those tools, configured for it to check when the file is open, and opened each file.
Notes: It doesn't recognize | as a text break; need to find how to turn off rules

PR #969, opened 2020-08-10
Replace lunr search with Algolia DocSearch

HarunD:


bmansheim:
Replaced with RediSearch

PR #972, opened 2020-08-14
Fixed absolute link to Download Center

amiramm:


amiramm:
Hopefully, adjust CI to catch these.

PR #976, opened 2020-08-19
Update module release notes

bmansheim:


K-Jo (approved):
let's also add 2.0.19 of graph

PR #980, opened 2020-08-22
Remove "IP address" from Participating Clusters section

kashish-redislabs:
IP addresses are not supported to create CRDB clusters and the UI rejects it for obvious reasons. The documentation mentions IP address which confuses the customers.
In addition, on the same page, in the Prerequisites section, it says the following which is contradicting to the example which includes IP address.

`Network connectivity and cluster FQDN name resolution between all participating clusters`


AlonMagrafta:
The PR is right, but we need to add some more information and clarifications. @bmansheim let's talk about it on Thursday

AlonMagrafta (approved):

bmansheim:
Replaced by https://github.com/RedisLabs/redislabs-docs/pull/1084 due to merge conflicts

PR #985, opened 2020-08-24
Updated proxy section

dallastexas92:
Original docs stated there could be more than one proxy per node - this isn't correct!


AlonMagrafta (approved):
Fix is right.
Maybe the original text was to emphasized we could have multiple threads but that came out confusing and impling multiple proxies.

PR #986, opened 2020-08-25
DOC-343 Specify on landing page that AI and Gears are RS only

bmansheim:
https://docs.redislabs.com/staging/DOC-343-specify-RC-support/modules/

The quick start pages state that you need RS to run AI and Gears so I think we just need to specify on the landing page that they are RS only.


K-Jo (approved):
Is this the only location though?  And is it clear to the user what RS is?  Isn't it "Software"?

PR #993, opened 2020-08-27
Update design of header and footer

HarunD:
This change updates the design of header and footer to match the new marketing website.


bmansheim:
@HarunD I made a few more tweaks. Only think left is the font. I see you added it and it shows up properly in a local build but not on the staging site. Any ideas?

HarunD:
> @HarunD I made a few more tweaks. Only think left is the font. I see you added it and it shows up properly in a local build but not on the staging site. Any ideas?

@bmansheim It was an issue with import path, it's fixed now.

PR #999, opened 2020-08-30
DOC-324  Update grafana json templates

bmansheim:


bmansheim:
Branched from redesign by mistake

PR #1002, opened 2020-08-31
Update backup instructions for GCP

yanivhershkovich:
Added clarifications for Google Cloud Storage setup as it's missing a lot of steps needed in order to configure it.

### DIFF
--- a/content/embeds/backup-locations.md
+++ b/content/embeds/backup-locations.md
@@ -104,7 +104,12 @@ Before you choose to backup to Google Cloud Storage, make sure that you have:
 - Private key
 
 {{< note >}}
-You can retrive the client and key details from a defined service account in GCP console (go to API & Services > Credentials > Create Credentials).
-The service account should be set with `Storage Legacy Bucket Writer` permissions on the target bucket. Verify the bucket doesn't use a retention policy as it can interfere with the process.
-Note that the format of the private key from the downloaded JSON is in a single string where new lines are marked with `\n` characters. When pasting the key into the UI, reformat the key by replacing all `\n` characters with a new line.
+
+You can find the client and key details in your service account in the GCP console (**API & Services** > **Credentials** > **Create Credentials**).
+
+- Make sure that the service account has the `Storage Legacy Bucket Writer` permission on the target bucket.
+- Make sure that the bucket doesn't use a retention policy because it can interfere with the process.
+- The format of the private key from the downloaded JSON is in a single string where new lines are marked with `\n` characters.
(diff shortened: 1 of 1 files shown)


AlonMagrafta:
@yanivhershkovich Thanks for the clarifications.
I guess the other cloud providers require similar steps, can you advise?
@bmansheim what do you say about inserting this in a collapsed + with a title.
Both - I'm conflicting since we are documenting here step on GCP. So a part of me thinks - this helps the user, we should have it, but another part of me says - it's not our responsibility to document the "how-to" on the GCP side. What happens if something changes on their side? What happens when a support ticket will arrive and say "hi, I followed your steps for GCP but something is not working for me there".
Again, I'm not against, just thinking out loud.
Let me know what you think

yanivhershkovich:
@AlonMagrafta 
The steps are specific to GCP. 
I'd also add another note that the bucket shouldn't use a retention policy as it can interfere with the process.

I do suggest adding it to our documentation even with the risk of it being changed in the future.
I think It looks worse when a customer asks how to configure it and we don't have/know the steps.
We had that bad experience with one customer and as a result, I submitted this request.

amiramm:
> @yanivhershkovich Thanks for the clarifications.
> I guess the other cloud providers require similar steps, can you advise?
> @bmansheim what do you say about inserting this in a collapsed + with a title.
> Both - I'm conflicting since we are documenting here step on GCP. So a part of me thinks - this helps the user, we should have it, but another part of me says - it's not our responsibility to document the "how-to" on the GCP side. What happens if something changes on their side? What happens when a support ticket will arrive and say "hi, I followed your steps for GCP but something is not working for me there".
> Again, I'm not against, just thinking out loud.
> Let me know what you think

Good point @AlonMagrafta - we need to focus on the subtleties of getting these to work with our solution - I think this PR mostly addresses that. I.e. if tomorrow GCP removes the line breaks or changes the format - no harm is done. We can recast this a little to highlight what is the format we expect to get, but overall, in this case, I think this is fine. @bmansheim @banker In more general terms, beyond keeping encouraging everyone using the docs to submit PRs when they see something, we should "regression test" the entire doc set whenever we have a major release.

AlonMagrafta:
@yanivhershkovich @amiramm Thanks for the feedback.
@bmansheim I did some rewrite and added Yaniv's last comment about the retention policy. Please review and let's approve it together

bmansheim:
@AlonMagrafta I made a few edits. Let me know if they are ok.

AlonMagrafta (approved):

PR #1006, opened 2020-09-02
running same OS all nodes

AlonMagrafta:


bmansheim:
- Move this to Supported Platforms article
- Add article about upgrading OS of nodes, either migrating data to new nodes or maintenance mode and upgrade

bmansheim:
Opened another ticket for instructions to upgrade OS on nodes

PR #1007, opened 2020-09-02
Added a quick start guide for generic K8s

alexmilowski:
A simple quick start guide.


content/platforms/kubernetes/quick-start.md
```diff
@@ -0,0 +1,164 @@
...
+When installed, the following are created:
+
+ * a service account under which the operator will run,
+ * a set of Roles that define the privileges necessary for the operator to perform its tasks,
+ * a set of Role Bindings to the service account to authorize the service account to have the correct roles (see above),
+ * the CRD for a Redis Enterprise Cluster,
+ * the CRD for a Redis Enterprise Database,
+ * the operator (a deployment).
+
+ The operator currently runs within single namespace and is scoped to operate only on cluster CRD's in that namespace. This allows a multi-tenant deployment of the operator supporting different versions in different namespaces.
```
amiramm, line 41:
```suggestion
 The operator currently runs within a single namespace and is scoped to operate only on Redis Enterprise Cluster CRD's in that namespace. This allows the deployment of the multiple instances of the operator on the same Kubernetes cluster, supporting multiple Redis Enterprise Clusters in multiple namespaces.
```
Multi-tenancy is ambiguous as it may refer to K8s cluster or Redis Enterprise Cluster. We need to find a way to promote the pattern of multiple DBs on the cluster while separately discussing the benefits or multiple Enterprise clusters... regarding versions - to a degree - in that the CRD at the K8s cluster level may limit which version you can deploy - see if you can work that into here or maybe at this level we keep it simple and just mention that the operator is namespace scoped.

content/platforms/kubernetes/quick-start.md
```diff
@@ -0,0 +1,164 @@
...
+
+You need to ensure you pull the correct version of the bundle. The version tags
+can be found by checking the [releases on Github](https://github.com/RedisLabs/redis-enterprise-k8s-docs/releases)
+or via the GitHub API.
+
+The bundle for the latest release can be downloaded via `curl` using the following:
+
+```
+
+VERSION=`curl --silent https://api.github.com/repos/RedisLabs/redis-enterprise-k8s-docs/releases/latest | jq -r .tag_nam`
```
amiramm, line 59: you're requiring folks to have `jq` installed on their system. Can this be more generic?

content/platforms/kubernetes/quick-start.md
```diff
@@ -0,0 +1,164 @@
...
+or you can replace `VERSION` in the above with a specific release tag you desire.
+
+
+If you inspect this YAML file, you will find definitions for the various items
+described in the previous section.
+
+You can install the operator with a single apply command:
+
+```
+kubectl apply -f bundle.yaml -n my-namespace
```
amiramm, line 72: one way to simplify this, as we do in other guides, is to change the context to the namespace you're operating in so you won't need to specify the namespace explicitly every time. WDYT? why did you choose to use `-n`?

content/platforms/kubernetes/quick-start.md
```diff
@@ -0,0 +1,164 @@
...
+that contains the specification of the cluster option. See the
+[Redis Enterprise Cluster API reference](https://github.com/RedisLabs/redis-enterprise-k8s-docs/blob/master/redis_enterprise_cluster_api.md)
+for more information on the various options available.
+
+You can test the operator by creating a minimal cluster by following this procedure:
+
+1. Create a file called `simple-cluster.yaml` that defines a Redis Enterprise
+   cluster with three nodes:
+
+    ```
```
amiramm, line 113: I think you should add a pre-requisites reference since this deployment requires at least 3 nodes with at least 4cpu/GB each whereas the Operator deployment is very modest.

content/platforms/kubernetes/quick-start.md
```diff
@@ -0,0 +1,164 @@
...
+   ```
+
+   or by simply looking at the status of all the resources in your namespace:
+
+   ```
+   kubectl get all -n my-namespace
+   ```
+
+4. Once the cluster is running, follow the information in [Managing Redis Enterprise Databases in Kubernetes]({{< relref "db-controller.md" >}})
+   to create databases.
```
amiramm, line 164: why not add this here and make it a "to a working database" quick-start? what most people want is a working endpoint(?)

amiramm (changes requested):
Looks good. see comments.

bmansheim:
@alexmilowski Do you want to respond to the comments?

content/platforms/kubernetes/quick-start.md
alexmilowski: curl ... | grep tag_name | awk  -F'"' '{print $4}'

content/platforms/kubernetes/quick-start.md
alexmilowski: It's the most generic from a cut-n-paste scenario ... if you presume a default namespace, people can't cut-n-paste a command without doing that first. consistency is better... I will add: kubectl config set-context --current --namespace=

alexmilowski: I have added a section on namespaces that clarifies this.

content/platforms/kubernetes/quick-start.md
alexmilowski: Added a note and example of configuring the node resources.

content/platforms/kubernetes/quick-start.md
alexmilowski: Good point - added.

content/platforms/kubernetes/getting-started/quick-start.md
```diff
@@ -0,0 +1,228 @@
...
+    cat <<EOF > simple-cluster.yaml
+    apiVersion: "app.redislabs.com/v1"
+    kind: "RedisEnterpriseCluster"
+    metadata:
+      name: "test-cluster"
+    spec:
+      nodes: 3
+      redisEnterpriseNodeResources:
+        limits:
+          cpu: 1000m
```
amiramm, line 153: I would not run on anything lower than 2 CPU and 3GB in a K8s environment and might be worthwhile to point to HW requirements doc since this isn't a configuration supported, even for a dev cluster.

amiramm (changes requested):
LGTM, minor comment. Plz consider ending with a connection to the database (for example, with crdb-cli)

alexmilowski:
This now works with the K8s reorg and I've updated it to address the comments.

content/platforms/kubernetes/getting-started/quick-start.md
```diff
@@ -0,0 +1,228 @@
...
+    apiVersion: "app.redislabs.com/v1"
+    kind: "RedisEnterpriseCluster"
+    metadata:
+      name: "test-cluster"
+    spec:
+      nodes: 3
+    EOF
+    ```
+
+    This will request Redis Enterprise nodes with 2 CPUs and 4GB of memory.
```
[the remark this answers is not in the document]
alexmilowski: Those are the defaults. I'll make this more clear.

content/platforms/kubernetes/getting-started/quick-start.md
alexmilowski: I do this all the time in my testing but I won't use that in an example in these docs. :)

amiramm (approved):
LGTM

bmansheim:
Merged to https://github.com/RedisLabs/redislabs-docs/pull/1174 for style review

PR #1018, opened 2020-09-10
Update add-module-to-database.md

kurtfm:
It is currently not possible to add a module to an existing database.  Acc to Pieter this is coming.


bmansheim:
Looks like the only instance of this issue.

PR #1019, opened 2020-09-12
Search experiment

abrookins:
Replaces the Lunr.js search form on the home page with a new RediSearch-backed search experience.

image

At this point, the front-end is pinned to a specific App Engine URL. We can use this for testing, but the end goal is to run the search app in three regions (US-West, Zurich, and Mumbai) and use a global load-balancer to route requests to the nearest region.


bmansheim:
@abrookins Really looking forward to this (for about 2 years...). Thanks for taking up the call!

static/js/search.js
```diff
@@ -1,176 +1,119 @@
-var lunrIndex, pagesIndex;
-
-function endsWith(str, suffix) {
-    return str.indexOf(suffix, str.length - suffix.length) !== -1;
+// const SEARCH_API_URL = "https://docsearch-dot-redislabs-university.appspot.com/search"
+const SEARCH_API_URL = "https://docsearch-dot-redislabs-university.appspot.com/search"
```
abrookins, line 2: Once we have a load balancer set up, we need to point this app at the URL that we load-balance across regions instead of a specific App Engine instance, like we do here.

abrookins:
@bmansheim Hey, Ben! A coworker looked this over, and we think it's ready for staging. 🚀

bmansheim:
Moved to https://github.com/RedisLabs/redislabs-docs/pull/1025 for review

PR #1021, opened 2020-09-14
Add Active-Active create step to verify participating clusters

yorozen:
need to verify the database was created on all participating cluster


bmansheim:
@yorozen You do that by opening the CRDB database configuration and seeing that all of the participating clusters are listed?

AlonMagrafta (approved):

PR #1035, opened 2020-09-21
RedisInsight: Added note about using RI with a reverse proxy.

nishantgeorge:
Screenshot 2020-09-21 at 4 48 46 PM

This note is relevant to all previous versions of RedisInsight as well. It can be merged after it is approved, no need to wait for the next release.

### DIFF
--- a/content/ri/using-redisinsight/troubleshooting.md
+++ b/content/ri/using-redisinsight/troubleshooting.md
@@ -32,5 +32,5 @@ We are happy to receive your feedback at redisinsight@redislabs.com.
 
 RedisInsight should work fine behind a reverse proxy like Nginx for most use-cases.
 There are a couple of things to keep in mind:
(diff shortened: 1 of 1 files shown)


slemeur:
Good ! I think we need to clarify where the timeout conifguration must be done

nishantgeorge:
> Good ! I think we need to clarify where the timeout conifguration must be done

The configuration has to be done in the reverse proxy. I thought that is obvious. If it's not, we can provide an example for Nginx. We can include this line:

"For example, for Nginx, the `proxy_read_timeout` directive should be set to `30s` or greater."

@slemeur Is that okay?

slemeur:
let's just mention that:
> we recommend that the **request timeout is set to over 30 seconds** on the reverse proxy.

slemeur (approved):

PR #1041, opened 2020-09-24
Update support for modules on Active-Active

yanivhershkovich:
Under 'Active-Active database current limitations' point 3 (i.e. Active-Active databases do not support Redis modules) is outdated - 
a) I think we already support RediGears & RediSearch 2.0 (and maybe ReJSON?)
b) The link to 'Redis modules' doesn't seem to work.

I didn't make changes, just added '>>> OUTDATED <<<' on the relevant line

### DIFF
--- a/content/rs/administering/designing-production/active-active.md
+++ b/content/rs/administering/designing-production/active-active.md
@@ -39,11 +39,15 @@ Before configuring an Active-Active database, you must:
 - Configure the network so that all nodes in each cluster can connect to the proxy port and the cluster admin port (9443) of each cluster.
 - Confirm that a [network time service](#network-time-service-ntp-or-chrony) is configured and running on each node in all clusters.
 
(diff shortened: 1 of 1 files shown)


amiramm:
@AlonMagrafta FYI - can you suggest language? 
@yanivhershkovich  - ReJSON - not yet. I believe RediSearch 2.0 only, but will let Alon clarify.

bmansheim:
(@AlonMagrafta Amplifying)
> AlonMagrafta  FYI - can you suggest language?
> yanivhershkovich - ReJSON - not yet. I believe RediSearch 2.0 only, but will let Alon clarify.

AlonMagrafta:
@bmansheim Didn't I do it with my commit? maybe it got missed :) 
Anyhow, yes, it's only Search 2.0 and Gears

PR #1049, opened 2020-09-30
Update using-oss-cluster-api.md

yorozen:
typo should be - and not _

Usage: crdb-cli crdb update [OPTIONS]

  Update configuration of an existing CRDB.

Options:
  --crdb-guid TEXT                Database GUID  [required]


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/1050

PR #918, opened 2020-07-10
Redis Enterprise Software Security

banker:

PR #1051, opened 2020-09-30
DOC-353 Add article for password rotation

bmansheim:


bmansheim:
@AlonMagrafta Please review.

AlonMagrafta:
Looks very good! Two thing that are important to mentions (one "note" the second "warning").
The Note -
RS users can access to the management API+UI and/or the DB connection - depending on their role.
When setting multiple passwords, it applies to all connections (API/UI and DB connections).

Warning - if you have a user with multiple passwords and you set a new password via the UI - it will override all the previous ones (-i.e. delete all the existing and add the new as a single new password).

AlonMagrafta (changes requested):
Looks very good! Two thing that are important to mentions (one "note" the second "warning").
The Note -
RS users can access to the management API+UI and/or the DB connection - depending on their role.
When setting multiple passwords, it applies to all connections (API/UI and DB connections).

Warning - if you have a user with multiple passwords and you set a new password via the UI - it will override all the previous ones (-i.e. delete all the existing and add the new as a single new password).

banker:
@bmansheim Please add this new page to https://github.com/RedisLabs/redislabs-docs/pull/918 rather than merging in at its current location

bmansheim:
@AlonMagrafta 
> The Note -
> RS users can access to the management API+UI and/or the DB connection - depending on their role.
> When setting multiple passwords, it applies to all connections (API/UI and DB connections).

Isn't this obvious? Is there a way to give a user a password for the UI/API and a different password for the data?

AlonMagrafta:
It is obvious to us, but the field hasn't been fully used to the two kinds of permissions and it is my feeling it something worth pointing out at this time :)

AlonMagrafta (approved):

PR #1060, opened 2020-10-08
Simplify the kubernetes intro

abrookins:


amiramm:
Except for that first paragraph, this isn't an actual published article in docs.

PR #1063, opened 2020-10-09
Cache the no results document for queries with 0 results

abrookins:
Currently, an empty result set is cached for queries with zero results. That means you'll see the 'no results' message once, but if you type the query a second time, you won't see the 'no results' message.

This change caches the fake document that powers the 'no results' message, so that you see it again if you type the same query.


bmansheim:
LGTM

PR #1070, opened 2020-10-14
DOC-355 Module dependency installation not required

bmansheim:


content/modules/redisgears/installing-redisgears.md
```diff
@@ -16,6 +16,10 @@ Before you can use RedisGears, you have to install the RedisGears module on your
 
 ### Step 1: Install RedisGears dependencies
 
+{{< note >}}
+This step is required for RS version 6.0.8 and below only.
```
banker, line 20: @bmansheim 

Can we do v6.0.8 instead of "version 6.0.8"? This is an important style decision we need to make.

Also, this page references RS all over the place without ever defining it. At the very least, pages like this should use the full product name at the beginning, followed by the abbreviation in parentheses.

banker: Ways to represent this:

RS v6.0.8
RS 6.0.8
Redis Enterprise Software 6.0.8
Redis Enterprise Software v6.0.8
Redis Enterprise Software version 6.0.8

banker: My preference:

This step is required for Redis Enterprise Software 6.0.8 and below.

What should we do @bmansheim ?

PR #1077, opened 2020-10-21
Fix FT.CREATE example and add example for document scoring

elena-kolevska:
Changing the score field name in the FT.CREATE example to something else than the reserved `WEIGHT`.
Currently, the following error is returned:

```
(error) Bad arguments for weight: Could not convert argument to expected type
```

Added an example for document scoring.

### DIFF
--- a/content/embeds/tryout-redisearch-2.md
+++ b/content/embeds/tryout-redisearch-2.md
@@ -1,25 +1,26 @@
 ### Creating indexes
 
-Let's create a new index called "database_idx".
+Let's create a new index called "`database_idx`".
 When you define the index, you must pass in the structure of the data you are adding to the index.
-In this example, we have four fields: title (TEXT), body (TEXT), url (TEXT), and weight (NUMERIC), and the title has a weight of 5.0.
+In this example, we have four fields: `title` (`TEXT`), `body` (`TEXT`), `url` (`TEXT`) and `visits` (`NUMERIC`), with the title field having a higher weight than the others (`5.0`).
 
-1. Connect to Redis.
+1. Connect to Redis (replace 12543 with the port number your RediSearch is running on).
 
     ```sh
     $ redis-cli -p 12543
     127.0.0.1:12543>
     ```
 
-1. Create the schema:
+2. Create the schema:
 
     ```sh
-    127.0.0.1:12543> FT.CREATE database_idx PREFIX 1 "doc:" SCORE_FIELD "doc_weight" SCHEMA title TEXT body TEXT url TEXT doc_weight NUMERIC
+    127.0.0.1:12543> FT.CREATE database_idx PREFIX 1 "doc:" SCORE_FIELD "doc_weight" SCHEMA title TEXT body TEXT url TEXT visits NUMERIC
     ```
 
-    This command indexes all of the hashes with the prefix "doc:".
-    By default, all hashes get a score of 1.0.
-    In this example, the SCORE_FIELD directive specifies a "weight" field whose value can override the default score for a document.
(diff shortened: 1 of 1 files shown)


K-Jo:
@elena-kolevska isn't it confusing that the field for the score is also indexed?

elena-kolevska:
@K-Jo That's how it was before, in the example, so I thought it has to be. If that's not the case, then yes, I'd say it's better to remove it.

K-Jo:
Can you please remove it from the schema?  Or add another field to the schema that isn't the doc change @elena-kolevska @bmansheim

elena-kolevska:
@K-Jo Ok, done. There were a few other small things that needed to be fixed, I corrected them too, as well as adding a sentence or two more on document scoring.

content/embeds/tryout-redisearch-2.md
```diff
@@ -1,25 +1,26 @@
...
-    127.0.0.1:12543> FT.CREATE database_idx PREFIX 1 "doc:" SCORE_FIELD "weight" SCHEMA title TEXT body TEXT url TEXT weight NUMERIC
+    127.0.0.1:12543> FT.CREATE database_idx PREFIX 1 "doc:" SCORE_FIELD "doc_weight" SCHEMA title TEXT body TEXT url TEXT visits NUMERIC
     ```
 
-    This command indexes all of the hashes with the prefix "doc:".
-    By default, all hashes get a score of 1.0.
-    In this example, the SCORE_FIELD directive specifies a "weight" field whose value can override the default score for a document.
+    This command indexes all of the hashes with the prefix "`doc:`" as well as all the future ones that will be created.
+
+    By default, all hashes get a score of 1.0, but if we need to be able to assign document-specific score to documents, we can do that too. We only need to specify in the schema the name of the hash element in which we will define the document score which will override the default of 1.0. This is done by using the `SCORE_FIELD` directive.
```
bmansheim, line 22:
```suggestion
    By default, all hashes get a score of 1.0, but if we need to assign a document-specific score to documents, we can do that too.
    We only need to specify in the schema the name of the hash element in which we will define the document score that overrides the default of 1.0. This is done by using the `SCORE_FIELD` directive.
```

content/embeds/tryout-redisearch-2.md
```diff
@@ -33,22 +34,28 @@ HSET database_idx {doc}:1 ...
...
-- Title: "Redis Labs"
-- Body: "primary and caching"
-- URL: <https://redislabs.com>
-- Weight: 10
+```sh
+127.0.0.1:12543> HSET doc:1 title "Redis Labs" body "Primary and caching" url "<https://redislabs.com/primary-caching>" visits 108
+OK
+```
 
+To add a document specific score, causing the document to appear higher or lower in results, set a value for the `doc_weight` field, which we specified as the `SCORE_FIELD` in the schema definition:
```
bmansheim, line 49:
```suggestion
To add a document specific score that causes the document to appear higher or lower in results, set a value for the `doc_weight` field.
We specified the `SCORE_FIELD` in the schema definition to hold the document weight value.
```

bmansheim (changes requested):

content/embeds/tryout-redisearch-2.md
elena-kolevska: Sorry @bmansheim , I had committed some other changes without seeing your request. Applied now.

bmansheim:
@K-Jo We're ready to go on this, right?

K-Jo (approved):

bmansheim (approved):

PR #1096, opened 2020-11-09
Improve Active-Active docker

bmansheim:


content/embeds/docker-memory-limitation.md
```diff
@@ -0,0 +1,3 @@
+If you cannot activate the database because of a memory limitation,
+make sure that Docker has enough memory allocated in the Advanced section
+of Docker Settings.
```
banker, line 3: Please use the appropriate convention to call out "Advanced" and "Docket Settings". This is a place where it'd be useful to link to the official Docker documentation.

content/rs/getting-started/getting-started-active-active.md
```diff
@@ -92,10 +92,10 @@ Now we have two Redis Enterprise Software clusters with FQDNs
     1. In the **participating clusters** list, add the address and admin credentials for:
         - `https://cluster1.local:9443` - the cluster you are currently connected to
         - `https://cluster2.local:9443` - the other cluster
-    <!-- Also in create-crdb.md -->
+
     1. In the **Database clustering** option, either:
 
-        - Make sure the Database clustering is enabled and select the number of shards
+        - Make sure that Database clustering is enabled and select the number of shards
```
banker, line 98: "Database clustering" needs to be in bold I think?

content/rs/getting-started/getting-started-active-active.md
```diff
@@ -92,10 +92,10 @@ Now we have two Redis Enterprise Software clusters with FQDNs
     1. In the **participating clusters** list, add the address and admin credentials for:
         - `https://cluster1.local:9443` - the cluster you are currently connected to
         - `https://cluster2.local:9443` - the other cluster
-    <!-- Also in create-crdb.md -->
+
     1. In the **Database clustering** option, either:
```
banker, line 96: "In" strikes me as the wrong preposition here. We usually don't "In the X option". Maybe "For"?

content/rs/getting-started/getting-started-active-active.md
```diff
@@ -109,13 +109,12 @@ Now we have two Redis Enterprise Software clusters with FQDNs
...
-    Note: If you cannot activate the database because of a memory limitation,
-    make sure that Docker has enough memory allocated in the Advanced section
-    of Docker Settings.
+    {{< note >}}
+    {{< embed-md "docker-memory-limitation.md" >}}
+    {{< /note >}}
 
-1. After the Active-Active database is created, you can now visit each cluster 1 at
-https://localhost:8443 and cluster 2 at https://localhost:8445.
+1. After the Active-Active database is created, access the RS admin console
```
banker, line 116: If there's no other admin console, why do we need to quality this with "RS"?

content/rs/getting-started/getting-started-active-active.md
bmansheim: This is a comment to indicate that the same content is used in another article. It doesn't show in the HTML.

content/embeds/docker-memory-limitation.md
bmansheim: Unfortunately it's different per OS. Do you think we should have links for each OS?
Currently it's written for Windows which itself is a problem.
I would lean toward making it more generic and that customers should go consult the Docker docs if they need to. I guess we can link to https://docs.docker.com/.

content/rs/getting-started/getting-started-active-active.md
bmansheim: I would treat it as a product name, but I can hear the argument against also.
Should it be Admin Console or admin console?

PR #1104, opened 2020-11-15
Custom installation with same user/group IDs

GitaShmueli:


bmansheim:
@GitaShmueli Acc to Gidon this needs verification.

content/rs/installing-upgrading/_index.md
```diff
@@ -170,6 +170,7 @@ sudo ./install.sh --install-dir <path> --config-dir <path> --var-dir <path>
 ### Custom installation user and group
 
 By default, RS is installed with the user:group `redislabs:redislabs`.
+If you require `redislabs` UID (User Identifier) and GID (Group Identifier) numbers to be identical on all the nodes, please create `redislabs` user and group prior the installation with the required numbers.
```
gidonmiller, line 173: the addition looks fine but it is under the custom installation section and IIUC this is not what was done by @GitaShmueli . this should also work with a custom install so maybe it needs to be outside of the custom installation section?

GitaShmueli: @bmansheim, Gidon is right, please tell me if I need to open another PR with the comment in the correct place.

bmansheim: @GitaShmueli That's OK. I'll take care of it.

PR #1108, opened 2020-11-18
Recover configuration for databases

GitaShmueli:
The option for recovering configuration appears only in the A-A description and missing for the regular DB section.
In addition, I think need to mention for A-A restore, to use the data recovery only on the cluster that has the most updated data.
Other participating clusters should be created with configuration only and sync from the cluster on which the data restored.
But need to discuss it with R&D


bmansheim:
Replaced by https://github.com/RedisLabs/redislabs-docs/pull/1126

PR #1120, opened 2020-12-01
Fix instructions for DNS setup for endpoints

gilbennun:
this command fails because it should use a unique name , internal.clustername.domain.com is not a unique name because clustername.domain.com already exists.
in addition we need to instruct customer to create ns record for the new unique domain name


kaitlynmichaelredislabs:
This issue is being tracked internally by RED-50175 and DOC-564

PR #1122, opened 2020-12-02
Fix ioredis and node_redis code fences

bmansheim:
Staging site: https://docs.redislabs.com/staging/fix-node-formatting


simonprickett (approved):
@bmansheim I added one small extra commit for the ioredis page that fixes a couple of instances of missing `<` and `>` in `<hostname>` for example.  Checked this over locally for both node_redis and ioredis pages, and the code blocks look good to me now.

PR #1123, opened 2020-12-02
DOC-405 Revamp getting started and installation

bmansheim:
This PR is to address a request from the field to have unified instructions that cover all steps up to creating a database.

Planning your RS deployment
Downloading the installation package
Installing RS on Linux
Set up the new cluster
Add more nodes to the cluster
Next steps

https://docs.redislabs.com/staging/revamp-getting-started/rs/installing-upgrading/


AlonMagrafta:
@orshwartzredislabs I'll love your input here too.
See Ben's link to staging , might help with having a full overview and compare to the prev page.

bmansheim:
@AlonMagrafta @orshwartzredislabs I'm in the process of making more changes based on feedback from SAs. Most helpful would be your thoughts on what information should not be included in this installation page, if any.

AlonMagrafta:
@bmansheim ack

lanceleonard (approved):
Approving so we can get these live.  

Additional changes to be handled separately.

PR #1128, opened 2020-12-04
Verify bdb_egress description

yanivhershkovich:
I think there's a typo in the description of 'bdb_egress_bytes' and 'bdb_egress_bytes_max'.
Since it's outgoing traffic, it would make sense that it's traffic from the DB and not to the DB.
Please verify first that my understanding is correct.


bmansheim:
Confirmed with Guy Cohen

PR #1130, opened 2020-12-07
Standardize support link; remove please

bmansheim:


banker (changes requested):
Hey Ben,

This looks good! The only thing is to use this page: https://redislabs.com/company/support/

1. support.redislabs.com is just a login from. It's super confusing and doesn't present the email options, for example.
2. Links like this should always being with `https`, not `http`

PR #1132, opened 2020-12-09
Replace sharding term in clustering article

bmansheim:


content/rc/concepts/clustering.md
```diff
@@ -72,10 +72,10 @@ are supported with the following limitations:
     and pipelining are supported with Redis Cloud cluster
     like if it were a non-cluster DB.
 
-## Changing the sharding policy
+## Changing the hashing policy
 
 The clustering configuration of a Redis Cloud instance can be
```
orshwartzredislabs, line 77: I think we can change the line (76) - "The clustering configuration of a Redis cloud instance.." 
To: 
"The hashing policy of a Redis Cloud instance can be changed" or "is configurable."

Then it will fit perfectly to the rest of the line that you modified - "However, hashing policy changes..."

content/rs/concepts/high-availability/clustering.md
```diff
@@ -31,18 +31,15 @@ In clustering, these terms are commonly used:
 - Slot or Hash Slot - The result of the hash calculation.
 - Shard - Redis process that is part of the Redis clustered database.
 
-## When to use sharding
+## When to use clustering
```
orshwartzredislabs, line 34: I think the right term here is actually sharding.
Sharding the data is the action of splitting the data into different Shards.

Also true for lines 36-37.


@AlonMagrafta ?

content/rs/concepts/high-availability/clustering.md
```diff
@@ -31,18 +31,15 @@ In clustering, these terms are commonly used:
...
+Clustering is an efficient way of scaling Redis that should be used when:
 
-- The dataset is large enough to benefit from using the RAM resources
-    of more than one server. We recommend sharding a dataset
-    once it reaches a size of 25 GB (50 GB for RoF).
-- The operations performed against the database are CPU-intensive,
-    resulting in performance degradation. By having multiple CPU cores
-    manage the database's shards, the load of operations is distributed
-    among them.
+- The dataset is large enough to benefit from using the RAM resources of more than one node.
```
orshwartzredislabs, line 38: Using the phrase - "enable clustering" is a little bit confusing because the user has no option to disable or enable the clustering in the cloud.

So maybe it can be rephrased to - 
"We recommend using the Redis Cluster technology to create multiple shards of...."

content/rs/concepts/terminology.md
```diff
@@ -33,7 +33,7 @@ database instances, referred to as "shards".
 RS supports various database configurations:
 
 - **Standard Redis database** - A single Redis shard with no
-    replication or data sharding.
+    replication or clustering.
```
orshwartzredislabs, line 36: I think the right term here is data sharding (the operation of splitting data into shards is sharding).
@AlonMagrafta ?

content/rc/concepts/clustering.md
AlonMagrafta: +1

content/rs/concepts/high-availability/clustering.md
AlonMagrafta: @bmansheim I know we discussed it and we assumed the term Clustering (the page title) was the one to use. It is even defined in the first paragraph of this page.
But, looking on https://redis.io/topics/cluster-tutorial I'm starting to think @orshwartzredislabs might be right.
Maybe we need to look at this as sharding (not clustering) and the operation of dividing the data to shards. And clustering to have multiple nodes per cluster.
WDYT?

content/rs/concepts/high-availability/clustering.md
AlonMagrafta: @orshwartzredislabs That's RS docs, so better be explicit about the steps. We can assume we are talking to customers using RS and not cloud users

content/rs/concepts/terminology.md
AlonMagrafta: same as above

amiramm: To add fuel to the fire - I think the right term is data partitioning... https://redis.io/topics/partitioning

AlonMagrafta: I was afraid to light this match ;)

bmansheim: Whether the correct term is clustering (as used in the database configuration page), sharding (as used informally and in the sharding policy), or partitioning (as in redis.io) we need to decide on the correct terms, define them, and use them correctly across the product.
I'll add this to the terminology doc.
cc: @banker

AlonMagrafta (approved):

PR #1138, opened 2020-12-14
Add REST API docs to the docs site

bmansheim:


lanceleonard:
This PR has been replaced by PR #1620, which takes a different approach.

PR #1142, opened 2020-12-16
It is important to specify what is the image for the Bootstrapper

laurentdroin:
Customers who need to specify the Bootstrapper image in their YAML need to understand what that image is.


alexmilowski (changes requested):
Okay but let's not put it in parentheses. 

**Operator and Bootstrapper**

bmansheim (approved):

PR #1149, opened 2020-12-20
Update to fix sample glitch in 'Using the API'

bmansheim:
Co-authored-by: Lance Leonard <lanceleonard.users.noreply.github.com>


bmansheim:
Another similar issue was reported and I found one more. Fixed all 3 in https://github.com/RedisLabs/redislabs-docs/pull/1157

PR #1155, opened 2020-12-22
Fix bad formatting of code snippet

TobyHFerguson:
Minor changes to correct code layout


bmansheim:
Fixed in https://github.com/RedisLabs/redislabs-docs/pull/1157

PR #1159, opened 2020-12-22
update 10-create-subscription.sh

TobyHFerguson:
Updated script because it use '||' as an 'or', which doesn't work for test conditions


bmansheim:
Unified in https://github.com/RedisLabs/redislabs-docs/pull/1156

PR #1160, opened 2020-12-23
DOC-377 Add k8s md files

bmansheim:


content/platforms/kubernetes/reference/GESHER.md
```diff
@@ -0,0 +1,255 @@
+---
```
amiramm, line 1: @alexmilowski @bmansheim This doc should move to Tasks section IMO. Too much for getting started and this is definitely not a reference doc. It needs some TLC and potentially breaking down but I suggest we do post it as is for now and take care of that later. Just not in the reference section.

content/platforms/kubernetes/reference/README.md
```diff
@@ -0,0 +1,107 @@
+---
```
amiramm, line 1: @alexmilowski @bmansheim This also seems like a Task section doc. And it needs a concept article as well and maybe an optional getting started section for the installation. WDYT?

content/platforms/kubernetes/reference/redis_enterprise_cluster_api.md
```diff
@@ -0,0 +1,180 @@
+---
+Title: Redis Enterprise Cluster API
```
amiramm, line 2:
```suggestion
Title: Redis Enterprise Cluster - Custom Resource Specification
```

content/platforms/kubernetes/reference/redis_enterprise_cluster_api.md
```diff
@@ -0,0 +1,180 @@
...
+description: 
+weight: 70
+alwaysopen: false
+categories: ["Platforms"]
+aliases:
+---
+
+This document describes the parameters for the Redis Enterprise Cluster custom resource.
+
+> Note this document is auto-generated from code comments. To contribute a change please change the code comments.
```
amiramm, line 12:
```suggestion
> Note this document is auto-generated from the Operator code.
```
This isn't actionable for the docs audience since the repo where you can contribute these changes is closed source. Any contributions here will have to be addressed in that repo.

content/platforms/kubernetes/reference/redis_enterprise_database_api.md
```diff
@@ -0,0 +1,268 @@
+---
+Title: Redis Enterprise Database API
```
amiramm, line 2:
```suggestion
Title: Redis Enterprise Database - Custom Resource Specification
```

content/platforms/kubernetes/reference/redis_enterprise_database_api.md
```diff
@@ -0,0 +1,268 @@
...
+description:
+weight: 80
+alwaysopen: false
+categories: ["Platforms"]
+aliases:
+---
+
+This document describes the parameters for the Redis Enterprise Database custom resource
+> Note this document is auto-generated from code comments. To contribute a change please change the code comments.
```
amiramm, line 11:
```suggestion
> Note this document is auto-generated from the Operator code. 
```

Same issue around contributing, as above.

amiramm (changes requested):
@bmansheim Thanks for this - please see my comments. @alexmilowski please review.

content/platforms/kubernetes/reference/GESHER.md
alexmilowski: I agree that this is a task.

alexmilowski: I don't think we should include this at all in this PR. While we need to add Gesher, the current docs on GitHub would need to be refactored to be used directly. Also, this is not reference material - more like a how-to guide (a task)

content/platforms/kubernetes/reference/README.md
alexmilowski: Similarly, we should not include this in this PR. This is not reference but a task/how-to guide. This content also should be refactored and not called "README"

content/platforms/kubernetes/reference/redis_enterprise_cluster_api.md
alexmilowski: This document is a more complete version of "cluster_options.md". We should not have both.

content/platforms/kubernetes/reference/redis_enterprise_database_api.md
alexmilowski: This document is a more complete version of db-options.md. We should not have both.

content/platforms/kubernetes/reference/README.md
alexmilowski: We should not include this document in this PR. It probably belongs in the tasks section and also overlaps with the GESHER.md document.

content/platforms/kubernetes/reference/GESHER.md
alexmilowski: We should not include this in this PR. This document belongs in the tasks section and also needs some refactoring.

content/platforms/kubernetes/reference/redis_enterprise_cluster_api.md
alexmilowski: This is a more complete version of cluster-options.md and we should not include both.

alexmilowski (changes requested):
The db-options.md & cluster-options.md files are hand authored documents that I wrote. They duplicate some of the information that is generated from the schemas for redis_enterprise_cluser_api.md and redis_enterprise_database_api.md. If we copy those across from the other GitHub project, we should delete the hand authored versions. Any edits we want we should work with the integrations team to make happen in the generated reference material.

Also, GESHER.md and README.md overlap each other (admission control). These need to be refactor and are also not reference material. These are tasks or "getting started" guides. We should not include those in this PR. Again, we should probably work with the integration team to refactor the documentation so they can be used here directly in the right place (probably as tasks).

kaitlynmichael:
Are these changes still relevant? Seems like they might be out of date considering all the conflicts with the current master branch. @banker @amiramm could you take a look?

amiramm:
> Are these changes still relevant? Seems like they might be out of date considering all the conflicts with the current master branch. @banker @amiramm could you take a look?
It seems like most of this is still relevant but this needs a deeper review and conflict resolution.

PR #1166, opened 2020-12-30
Update instructions in "Backing up to Azure Blob Storage (ABS)"

ecahana:
Update instructions in "Backing up to Azure Blob Storage (ABS)"

### DIFF
--- a/content/rc/administration/configuration/backups.md
+++ b/content/rc/administration/configuration/backups.md
@@ -117,8 +117,13 @@ Where:
 - *storage_account_access_key:* the primary access key to the
     storage account
 - *storage_account_name:* the storage account name
-- *container_name:* the name of the container, if needed.
+- *container_name:* the name of the container
 - *path*: the backups path, if needed.
+
+For example:
+URL: https://my.blob.core.windows.net/container
+Key: mysecretkey
+`abs://:mysecretkey@my/container`
 
 ## Backing up to an FTP Server
 
--- a/content/rc/administration/configuration/backups.md
+++ b/content/rc/administration/configuration/backups.md
@@ -86,22 +86,26 @@ To use an ABS container for storing your resources' backups, follow
 these steps in your [Microsoft Azure Management
 Portal](https://manage.windowsazure.com/):
 
-1. Access your storage by clicking the left-hand **STORAGE** icon.
-1. Select the storage account:
-    1. To create a new storage account:
-        1. Click the **NEW** button at the lower-left corner of the
-            page.
-        1. Verify that you've selected **DATA
-            SERVICES->STORAGE->QUICK CREATE** from the menu.
-        1. Enter the **URL** for your new storage account.
-        1. Select a **LOCATION/AFFINITY GROUP** for the storage
-            account.
-        1. Choose a **REPLICATION** mode for the account.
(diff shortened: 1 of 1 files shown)


content/rc/administration/configuration/backups.md
```diff
@@ -86,22 +86,26 @@ To use an ABS container for storing your resources' backups, follow
...
+2. Select the desired Storage Account
+3. Click on "Access Keys" on the left panel
+4. Click on "Show Keys" and copy the desired key
+
+Create Container:
+1. Click on "Storage Accounts" in Azure Console
+2. Select the desired Storage Account
+3. Click on "Containers" on the left panel
+4. Click on "+ Container"
+5. Give the container a name and click on "Create"
```
amiramm, line 108:
```suggestion
5. Give the container a name and click on "Create" 
6. Go to "Properties" and copy the URL
```

content/rc/administration/configuration/backups.md
```diff
@@ -86,22 +86,26 @@ To use an ABS container for storing your resources' backups, follow
 these steps in your [Microsoft Azure Management
 Portal](https://manage.windowsazure.com/):
```
amiramm, line 87:
```suggestion
Portal](https://portal.azure.com/):
```

content/rc/administration/configuration/backups.md
```diff
@@ -86,22 +86,26 @@ To use an ABS container for storing your resources' backups, follow
...
+
+Create Container:
+1. Click on "Storage Accounts" in Azure Console
+2. Select the desired Storage Account
+3. Click on "Containers" on the left panel
+4. Click on "+ Container"
+5. Give the container a name and click on "Create" 
 
 Set your resource's **Backup Path** to the path of your ABS storage
 account and clicking the **Apply** button using the following syntax:
```
amiramm, line 111:
```suggestion
account and clicking the **Apply** button using the following syntax, constructed from the key and URL you've copied earlier:
```

amiramm (changes requested):
LGTM, please see my changes and example commit.

content/rc/administration/configuration/backups.md
ecahana: re "constructed from the key and URL you've copied earlier:" - 
backup path should be abs://:storage_account_access_key@storage_account_name/container_name/[path/]
container URL for example - https://storageaccount123.blob.core.windows.net/storageaccount123container
the needed information can be extracted from the URL, but i'm not sure its more clear for the user than the current simple explanation in the doc (below):
storage_account_access_key: the primary access key to the storage account
storage_account_name: the storage account name
container_name: the name of the container, if needed.
path: the backups path, if needed.

kaitlynmichaelredislabs: @ecahana you believe we should leave line 111 as is, instead of adding "constructed from the key and URL you've copied earlier:", correct?

lanceleonard:
Hey all.  The underlying article was updated in a different PR.  It was updated to reflect current Azure UX, however, the specific details are handled by linking to Azure's docs on the subject (which are, strictly speaking, the authoritative resources on the subject).

Consequently, I'm not sure we need this PR any longer.  If there aren't any objections, I'm planning to close this.

lanceleonard:
Closing PR, per previous comment.

PR #1183, opened 2021-01-11
RedisInsight: Document ability to add database using GET URL

GnaneshKunal:
@slemeur @nishantgeorge 

**Do not merge without explicit approval.**

image


content/ri/using-redisinsight/api-get.md
```diff
@@ -0,0 +1,69 @@
...
+date: 2021-01-11 00:00:00 +0000
+weight: 16
+categories: ["RI"]
+path: api/
+nextStep:
+    Title: Memory Analysis
+    href: /docs/features/memory-analysis/
+---
+
+If you want to automate adding a Redis database without filling the [database form]({{< relref "/ri/using-redisinsight/add-instance.md" >}}), you might want to add database via GET URL.
```
kaitlynmichaelredislabs, line 12: When you say "via GET URL", where are you running this `GET` command?

content/ri/using-redisinsight/api-get.md
```diff
@@ -0,0 +1,69 @@
...
+Used to add Redis databases to RedisInsight.
+
+**URL** : `/add/`
+
+**Method** : `GET`
+
+
+### Query Parameters
+
+These are the required query parameters for any type of database.
```
kaitlynmichaelredislabs, line 27: Is this the order they need to be in as well?

content/ri/using-redisinsight/api-get.md
```diff
@@ -0,0 +1,69 @@
...
+
+| Parameter        | Type    | Required/Optional | Description                                                                  |
+|------------------|---------|-------------------|------------------------------------------------------------------------------|
+| `host`           | string  | required          | Hostname of your Redis database.                                             |
+| `port`           | number  | required          | Port of your Redis database.                                                 |
+| `name`           | string  | optional          | A nick name for the Redis database. Any string is valid                      |
+| `username`       | string  | optional          | Username of your Redis database.                                             |
+| `passsword`      | string  | optional          | Password of your Redis database.                                             |
+| `tls`            | boolean | optional          | `"true"` if your Redis database has TLS enabled.                             |
```
kaitlynmichaelredislabs, line 36: Why is this an option if the note below says you need to use the database form if you add TLS certs?

content/ri/using-redisinsight/api-get.md
```diff
@@ -0,0 +1,69 @@
...
+| `masterPassword` | string  | optional          | master password of the monitoring database                                   |
+| `redirect`       | boolean | optional          | `"true"` if you want to redirect to instance page after successful addition. |
+|                  |         |                   |                                                                              |
+
+{{< note >}}
+If you want to add the TLS certificates for your Redis database, you have to manually fill [database form]({{< relref "/ri/using-redisinsight/add-instance.md" >}}).
+{{< /note >}}
+
+
+#### Examples
```
kaitlynmichaelredislabs, line 49: Could we just give examples of the URL text rather than an entire screenshot?

kaitlynmichaelredislabs (changes requested):
I have several questions about this new procedure before I can add it.

content/ri/using-redisinsight/api-get.md
nishantgeorge: Not sure whether you mean the [Redis GET command](https://redis.io/commands/get). This has nothing to with that.

Any URL you hit by entering the URL in your browser's address bar is an HTTP GET request. So "via GET URL" basically translates to "by visiting this URL in your browser".

content/ri/using-redisinsight/api-get.md
nishantgeorge: No, the order of the parameters does not matter.

content/ri/using-redisinsight/api-get.md
nishantgeorge: A Redis database can have TLS enabled without requiring TLS certificates to be added.
The most simple case is that the Redis database simply has TLS enabled.
Then there are other more advanced configurations that would require TLS certificates to be added (in case client authentication or a custom CA certificate is required).

Screenshot 2021-02-05 at 1 42 29 PM

The `tls` parameter in the URL spec is only for the option marked `1`, and not the following options marked `2`. If those options need to be configured, the user would have to use the UI to add their Redis database.

content/ri/using-redisinsight/api-get.md
nishantgeorge: I think that should be fine. @GnaneshKunal Can you make the necessary changes, please?

content/ri/using-redisinsight/api-get.md
nishantgeorge: In the image above, the "Verify TLS Certificate" is supported with the `verifyServer` parameter.

content/ri/using-redisinsight/api-get.md
GnaneshKunal: Does the pattern of the URL looks fine or can I change the pattern `http://<redisinsight-host>:<redisinsight-port>/add/....`?

image

kaitlynmichaelredislabs: I think we definitely need to differentiate what is a variable and what is not. Maybe we could have two URLs, one using variable names and an example with sample values? That way the user can see both which values they need to supply and what they look like in context. 

http://add/?name=<redisinsight-host>&host=<local-host>&port=<redisdb-prt>
http://add/?name=mydb&host=172.19.02&port=6379

content/ri/using-redisinsight/api-get.md
kaitlynmichaelredislabs: So if the users are just typing this into their browser, how do they automate it? Is the URL something they can include in a program or script?

nishantgeorge: This is not meant for automation. It's useful in a classroom setting where the instructor can just share a GET URL with the class and the students can just follow the link to add the Redis database.
For proper automation using scripts, the [documented API](https://docs.redislabs.com/latest/ri/using-redisinsight/api/) should be used.

kaitlynmichaelredislabs: Are there any other use cases for this method besides a classroom setting?

nishantgeorge: I suppose it's any case where you want to add the Redis database easily, for example, a devops team.

kaitlynmichael (approved):

GnaneshKunal:
@kaitlynmichael Why haven't we merged this yet?

kaitlynmichael:
> @kaitlynmichael Why haven't we merged this yet?

The original description says not to merge without permission, so I was waiting for another approving review. Are we okay to merge this?

GnaneshKunal:
@kaitlynmichael 

> Are we okay to merge this?

Yeah.

PR #1189, opened 2021-01-14
api not working on 6.0.12

gilbennun:
when I use "-X POST" I get the error "method not allowed"
when using "/v1/cluster" I get the error "bad request"  (are you sure the API should be /v1/cluster and not /v1/users/password)


banker:
@gilbennun Can you please open a support ticket for this? https://redislabs.com/company/support/

amiramm:
@gilbennun @banker You're right, I'll fix the doc in a separate PR

PR #1194, opened 2021-01-15
Updated to the correct REST API endpoint

amiramm:


bmansheim (approved):
Verified on live system

Issue #261, opened 2019-03-11
Command segments need formatting


bmansheim:
**Expected behavior** (A concise description of what you found)
There are command segments in the text on https://docs.redislabs.com/latest/rs/administering/security/client-connections/ that aren't formatted as code.

**Expected behavior** (A concise description of what you expected to find)
All command and code segments should be in `backticks`.

**Screenshots** (If applicable, add screenshots of the problem)

**Additional context** (Add any other information)

ayruslore:
Opened a pull request to resolve the issue.
Pull Request Link - https://github.com/RedisLabs/redislabs-docs/pull/1197

bmansheim:
Good form, @ayruslore . Thank you for jumping in to help us with this.

PR #1197, opened 2021-01-16
Fix : Command segments formatting

ayruslore:
Issue 261 (https://github.com/RedisLabs/redislabs-docs/issues/261) - I have added back ticks for all the commands where it needed.


bmansheim:
@kaitlynmichaelredislabs @lanceleonard @banker What do you think about this formatting? We should probably make this explicit in our style guide.

kaitlynmichaelredislabs (approved):

PR #1202, opened 2021-01-20
DOC-513 Add -X POST to password rotation command

AlonMagrafta:


bmansheim:
@AlonMagrafta I added a note about running the command without '-X PUT' so that customers know that the command will behave differently if they leave that flag out. OK?

AlonMagrafta:
LGTM @bmansheim

PR #1220, opened 2021-02-02
Group search results by root page

abrookins:
Previously, search results rendered as a single list, with the root page duplicated, like this:

Screen Shot 2021-02-01 at 7 40 30 PM

Now, the search form groups results by their root page.

Grouped search

Staging link: https://docs.redislabs.com/staging/doc-search-structured/


kaitlynmichaelredislabs (approved):
I think this is an amazing improvement!

PR #1248, opened 2021-02-22
Update to upgrade procedure

amiramm:
Introduced pulling a version tag rather than cloning the repo.
TODO: add a user check for valid upgrade paths

### DIFF
--- a/content/platforms/kubernetes/tasks/upgrading-with-operator.md
+++ b/content/platforms/kubernetes/tasks/upgrading-with-operator.md
@@ -22,6 +22,11 @@ You can download the bundle for the latest release by issuing the following `cur
 VERSION=`curl --silent https://api.github.com/repos/RedisLabs/redis-enterprise-k8s-docs/releases/latest | grep tag_name | awk -F'"' '{print $4}'`
 curl --silent -O https://raw.githubusercontent.com/RedisLabs/redis-enterprise-k8s-docs/$VERSION/bundle.yaml
 ```
+For openshift environments, the name of the bundle is openshift.bundle.yaml, and so the curl command to run is:
+
+```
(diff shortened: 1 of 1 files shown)


laurentdroin:
Do we not need to say anything about Openshift? In case (I don't know if it's likely) that something specific to the openshift bundle changes, we would want the customers to apply the openshift bundle, no?

Also, it might be the rendering in github causing this but I see some formatting issues:

Screen Shot 2021-02-22 at 7 06 20 AM

Screen Shot 2021-02-22 at 7 09 21 AM

lanceleonard:
@laurentdroin - Use [this link](https://docs.redislabs.com/staging/amiramm-patch-1/platforms/kubernetes/tasks/upgrading-with-operator/) to view this doc on the staging server.  (Staging links only work until a PR is merged or closed.)

kaitlynmichael (approved):

kaitlynmichael:
@laurentdroin @amiramm I reformatted the note, so now it renders correctly. Is this PR ready to merge?

laurentdroin:
@amiramm I still have my initial concern - shouldn't we mention openshift?

amiramm:
> @amiramm I still have my initial concern - shouldn't we mention openshift?

@laurentdroin can you please suggest some language for that?

laurentdroin:
Reviewing all this, there are only two places that we need to amend for openshift and these are the actual commands we run to first download the bundle, and then apply the bundle.
So I guess that we should probably say something like "if you are using Openshift, run this instead:", twice. One time for each command.
We need to be careful not to make Openshift seems like a 2nd class citizen because it is actually more prevalent than non-OIpenshift, at least with all the customers I work with.

laurentdroin (changes requested):
Updated the file but not sure if my change worked.

kaitlynmichael (approved):

kaitlynmichael (approved):

PR #1258, opened 2021-03-04
rladmin cluster config saslauthd_ldap_conf requires an absolute path

laurentdroin:

PR #1259, opened 2021-03-04
mention path in rladmin command

vikrammoule-redislabs:
without mentioning the absolute path the command results in an error `ERROR: invalid token 'saslauthd.conf'`

### DIFF
--- a/content/platforms/kubernetes/tasks/ldap-on-k8s.md
+++ b/content/platforms/kubernetes/tasks/ldap-on-k8s.md
@@ -159,10 +159,10 @@ ldap_password: admin
 
 ### Using rladmin to update the configuration
 
-If this information was in a file called “ldap.conf”, you can connect to a Redis Enterprise Node pod and configure the server via:
+If this information was in a file called “ldap.conf” stored in "/tmp", you can connect to a Redis Enterprise Node pod and configure the server via:
 
 ```sh
-rladmin cluster config saslauthd_ldap_conf ldap.conf
+rladmin cluster config saslauthd_ldap_conf /tmp/ldap.conf
 ```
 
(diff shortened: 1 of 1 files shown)


content/platforms/kubernetes/tasks/ldap-on-k8s.md
```diff
@@ -159,10 +159,10 @@ ldap_password: admin
 
 ### Using rladmin to update the configuration
 
-If this information was in a file called “ldap.conf”, you can connect to a Redis Enterprise Node pod and configure the server via:
+If this information was in a file called “ldap.conf” stored in "/tmp", you can connect to a Redis Enterprise Node pod and configure the server via:
 
 ```sh
-rladmin cluster config saslauthd_ldap_conf ldap.conf
+rladmin cluster config saslauthd_ldap_conf /tmp/ldap.conf
```
kaitlynmichaelredislabs, line 165: We just merged a PR from @laurentdroin editing this file to `/path/to/ldap.conf`. Should it be specifically mentioned that this is stored in `/tmp` instead?

kaitlynmichaelredislabs (commented):
This PR edits the same lines as #1258 , we need to resolve any conflicts or close this as a duplicate.

content/platforms/kubernetes/tasks/ldap-on-k8s.md
laurentdroin: No, I don't think there is any assumption that the file would be in /tmp.
The lines above say:

```
If this information was in a file called “ldap.conf”, you can connect to a Redis Enterprise Node pod and configure the server via:

rladmin cluster config saslauthd_ldap_conf ldap.conf
```

So, no assumption where the file is (IMO).

kaitlynmichaelredislabs (approved):
Approved via discussion in slack #redislabs-docs

PR #1278, opened 2021-03-17
JIRA DOC-624: Block spiders access to out-of-date content

lanceleonard:
Removing out-of-date content from spiders.


banker (approved):
@lanceleonard I did some testing with the Google robots.txt tool, and I think this is what we're looking for.

PR #1292, opened 2021-03-30
ACLs-on-Active-Active

AlonMagrafta:
I thought it is worth mentioning :) 
Not urgent, and can be merged to master when ready

### DIFF
--- a/content/rs/security/passwords-users-roles.md
+++ b/content/rs/security/passwords-users-roles.md
@@ -128,4 +128,4 @@ We recommend that you disable the default user when using ACLs with your databas
 
 ### Users, Roles and Redis ACLs on Active-Active databases
 Users, Roles and Redis ACLs are cluster level entities. Therefore they are applied per a local participating cluster and Active-Active database instance and they are not replicated or propagated to the other participating clusters and instances.
-ACLs will be enforced according to the instace the client has connected to. The Active-Active replication mechnism will propagate all the effects of an operation.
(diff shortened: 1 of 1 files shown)


content/rs/security/passwords-users-roles.md
```diff
@@ -125,3 +125,7 @@ We recommend that you disable the default user when using ACLs with your databas
 {{< /note >}}
 
 ![default](/images/rs/default-user.png#no-click "default")
+
+### Users, Roles and Redis ACLs on Active-Active databases
+Users, Roles and Redis ACLs are cluster level entities. Therefore they are applied per a local participating cluster and Active-Active database instance and they are not replicated or propagated to the other participating clusters and instances.
```
kaitlynmichael, line 130:
```suggestion
Users, roles, and Redis ACLs are cluster-level entities. They are applied to a local participating cluster and Active-Active database instance, and are not replicated or propagated to the other participating clusters and instances.
```

content/rs/security/passwords-users-roles.md
```diff
@@ -125,3 +125,7 @@ We recommend that you disable the default user when using ACLs with your databas
...
+ACLs will be enforced according to the instace the client has connected to. The Active-Active replication mechnism will propagate all the effects of an operation.
```
kaitlynmichael, line 131:
```suggestion
ACLs will be enforced according to the instance the client is connected to. The Active-Active replication mechanism will propagate all the effects of the operation.
```

kaitlynmichael (changes requested):

kaitlynmichael (approved):

PR #1305, opened 2021-04-11
Strictly write the procedure to upgrade CRDB's

mikizuv2:
Some CRDB customers just miss and important rule to remember: Typical upgrade for a non-CRDB cluster is to first install.sh and then rladmin upgrade DB's.
For CRDB clusters you must first install.sh all nodes on all clusters and only then rladmin upgrade db (you must not rladmin upgrade DB before all clusters are upgraded via install.sh).

### DIFF
--- a/content/rs/installing-upgrading/upgrading.md
+++ b/content/rs/installing-upgrading/upgrading.md
@@ -9,7 +9,7 @@ aliases: /rs/administering/installing-upgrading/upgrading/
 To upgrade the Redis Enterprise Software (RS) software on a cluster,
 you must upgrade each of the nodes and then upgrade each of the databases in the cluster.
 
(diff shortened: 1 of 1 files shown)


content/rs/installing-upgrading/upgrading.md
```diff
@@ -9,6 +9,8 @@ aliases: /rs/administering/installing-upgrading/upgrading/
 To upgrade the Redis Enterprise Software (RS) software on a cluster,
 you must upgrade each of the nodes and then upgrade each of the databases in the cluster.
 
+For active active clusters you must upgrade all the nodes on all clusters first and then upgrade each of the databases in each cluster.
```
kaitlynmichael, line 12:
```suggestion
For Active-Active clusters, you must upgrade all the nodes on all clusters first, and then upgrade each of the databases in each cluster.
```

kaitlynmichael (approved):

PR #1310, opened 2021-04-12
Jira DOC-407: RC Reorg the REST API docs.

lanceleonard:
This work item simplifies the navigation of the Redis Cloud REST API docs.  Light edits for style (and clarity) have been made, but this is more along the lines of a quick copyedit, rather than an extensive dev edit (planned for a future work item.)

This commit is the initial check-in; review pass in progress.

Additional commits are expected.


kaitlynmichael (approved):
When I went to the staging link, there seems to be an empty "Concepts" page in the API section. Will that be removed?

lanceleonard:
> When I went to the staging link, there seems to be an empty "Concepts" page in the API section. Will that be removed?

That's the plan.  I'll check it out.  Thanks!

kaitlynmichael (changes requested):
Looks like that empty "Concepts" page is still there. Otherwise it looks good.

PR #1231, opened 2021-02-04
Rough draft of updated customize article.

lanceleonard:
This is an in-progress update.  Screen shots need to be added and additional content changes are expected.

PR #1312, opened 2021-04-13
Updated the python package to redis

charles2588:
Updated the python package to redis from redis-client as redis-client no longer works.

redis-client is no longer available on pypi.

```
sudo pip install redis-client
Password:
ERROR: Could not find a version that satisfies the requirement redis-client
ERROR: No matching distribution found for redis-client

```

Updated the sample python client code to include the redis pypi package instead.

### DIFF
--- a/content/rc/rc-quickstart.md
+++ b/content/rc/rc-quickstart.md
@@ -144,13 +144,13 @@ Here's how to connect to your database using a client for Python.
 1.  If you don't already have the client installed:
 
     ```sh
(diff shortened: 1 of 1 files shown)


lanceleonard:
Hi @charles2588; thanks for the contribution!  

In this case, we've updated the Python library in a [different PR](https://github.com/RedisLabs/redislabs-docs/pull/1231).

We appreciate the help, though.  We hope you'll continue to contribute!

PR #1313, opened 2021-04-13
Fix incorrect info about rsnap in memory analysis section. 

GnaneshKunal:
The rsnap doesn't actually leave the system/server where RI installed.


GnaneshKunal:
@K-Jo

Please review

lanceleonard (approved):

PR #1321, opened 2021-04-14
DOC-673: Broken images in RS: LDAP docs

lanceleonard:
Fixing image paths.


kaitlynmichael (approved):
For some reason on my screen most of these are looking really small, but otherwise perfect.

lanceleonard:
> For some reason on my screen most of these are looking really small, but otherwise perfect.

Yeah; it's something we'll need to check into.  (Still trying to figure out the right combination of settings/process for good screenshots on both Mac and Windows.)

PR #1352, opened 2021-04-29
To remove an unreachable instance, we need to use the --force flag

laurentdroin:
The doc incorrectly suggested using --no-force


lanceleonard (approved):
Looks good to me

PR #1355, opened 2021-05-03
Update password-rotation.md

yanivhershkovich:
Few issues on this page:
1. It doesn't seem like it will work on the database default access so there's a pre-requisite of adding custom users through ACLs to the database before you can add/rotate passwords. We should either mention that condition as a note/comment or include steps to a user/redirect to the ACLs page.
2. The API commands are broken to new lines and if a customer copies them as is, they won't work. It has to be a single line (I did the modification).
3. Fixed a minor typo ('an user' to 'a user' in two places).
4. Document title is 'Database Password Rotation' but it actually applies to all users, UI and Database and not only Database. We should consider renaming it but this is minor.

### DIFF
--- a/content/rs/administering/access-control/password-rotation.md
+++ b/content/rs/administering/access-control/password-rotation.md
@@ -1,5 +1,5 @@
 ---
-Title: Database password rotation
+Title: User password rotation
 description:
 weight: $weight
 alwaysopen: false
--- a/content/rs/administering/access-control/password-rotation.md
+++ b/content/rs/administering/access-control/password-rotation.md
@@ -1,30 +1,35 @@
 ---
-Title: Database Password Rotation
+Title: Database password rotation
 description:
 weight: $weight
 alwaysopen: false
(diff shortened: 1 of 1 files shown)


kaitlynmichael:
@yanivhershkovich this section is currently under "Administration" but I think it would be better under the "Security" section. Do you agree?

yanivhershkovich:
@kaitlynmichael I don't think so since this is about DB administration (i.e. password rotation) but if the PMs think it should move to security, I have no objection as long as we fix the issues in the page.

kaitlynmichael:
@yanivhershkovich do you think "User password rotation" would be a better title?

yanivhershkovich:
Yes, it might be better since it applies to both UI and DB users but I'll let the PMs decide :)

AlonMagrafta (approved):

AlonMagrafta:
@yanivhershkovich Thanks!

PR #1357, opened 2021-05-05
Publishing policy file changes

TobyHFerguson:
These changes have two effects:

1) To change the name of the S3 bucket as requested by SA management
2) To fully automate the publishing of the policy files and push them out to the S3 bucket (along with some processing) whenever they're changed. 
3) Added some clarifying documentation about how to use the cloudformation script, and updated the URLs to point to the new bucket
4) Added some clarifying documentation about how to use the terraform module, updated the URL to point to the new bucket, and corrected some errors in the terraform module


lanceleonard (approved):
Looks good to me.  (Note that we'll need to do an edit pass soon, but that's later.)

PR #1369, opened 2021-05-09
Add remark for envoy replacement

yorozen:
We Need to add a note or remark. On upgrade from version 5.X to 6.X, the default web server application was changed from nginx to envoy.
If any software is monitoring or allowing nginx, it should be updated to allow envoy.
We detected sp,e  software that was blocking envoy to listen to the required ports, such as TrendMicro Deep security,  dynatrace, and more.


yorozen:
https://redislabs.zendesk.com/agent/tickets/73472
https://redislabs.zendesk.com/agent/tickets/72183
https://redislabs.zendesk.com/agent/tickets/70475

PR #1382, opened 2021-05-26
Jira DOC 580: Update Database articles for new UX

lanceleonard:


kaitlynmichael (approved):
I really like the use of tables here.

PR #1385, opened 2021-05-28
Update redisgraph-2.4-release-notes.md

yanivhershkovich:
RedisGraph 2.4 requires cluster version 6.0.8 and above otherwise the module upload to the cluster fails.
This should be added as a requirement in the RN so customers are aware of it and are not surprised when the upload fails.

I'm not sure where to add it to the page so I just added <<Need to include cluste version requirement>> at the bottom of the page.
Maybe a 'Pre-Requisites` section?

I think this applies in general to every module RN, especially if it has such requirements.


kaitlynmichael:
> I think this applies in general to every module RN, especially if it has such requirements.

@yanivhershkovich Do you know where I could find these requirements for each module?

yanivhershkovich:
@kaitlynmichael I don't know if we have it documented.
I would think that Pieter Cailliau can be the best resource for that information or at least know from where to get that information.

kaitlynmichael:
> @kaitlynmichael I don't know if we have it documented.
> I would think that Pieter Cailliau can be the best resource for that information or at least know from where to get that information.

@K-Jo could you provide some guidance on what the cluster version requirements are and where would be best to document that?

lanceleonard:
@yanivhershkovich - This has been done as part of [PR #1692](https://github.com/RedisLabs/redislabs-docs/pull/1692).

Closing this PR in favor of that one.

PR #1398, opened 2021-06-10
DOC-737 K8s 6.0.20-4 release notes

kaitlynmichael:
DOC-737 K8s release notes for 6.0.20-4 
pulled from: https://github.com/RedisLabs/redis-enterprise-k8s-docs/releases/tag/v6.0.20-4

### DIFF
--- a/content/platforms/release-notes/k8s-6-0-20-4.md
+++ b/content/platforms/release-notes/k8s-6-0-20-4.md
@@ -1,0 +1,60 @@
+---
+Title: Redis Enterprise for Kubernetes Release Notes 6.0.8-20 (December 2020)
+description: Release notes for version 6.0.8-20
+weight: 72
+alwaysopen: false
+categories: ["Platforms"]
+---
+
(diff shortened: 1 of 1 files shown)


laurentdroin (commented):
We're missing the following, is it on purpose?

Compatibility Notes
OpenShift 4.7 and Rancher/kOps 1.20 are now supported
OpenShift 4.1, 4.2, 4.3 (previously deprecated) are no longer supported
GKE K8s version 1.14 (previously deprecated) is no longer supported
kOps (upstream K8s) 1.13, 1.14 (previously deprecated) are no longer supported

Deprecation notice
OpenShift 4.4 (no longer supported by Red Hat) is deprecated
GKE K8s versions 1.15, 1.16 (no longer supported by Google) are deprecated
kOps (upstream K8s) 1.15 is deprecated

laurentdroin:
Also, in order to stay consistent with the previous RN, don't we want to keep the title for each known limitation? The titles seem to have been stripped from the source github page.

kaitlynmichael:
I understand that we want to be consistent but I didn't think the Jira numbers and titles were adding any value on the public site. Internally, they are useful. For the public , I believe just a description of the problem was all that's necessary. If this has a significant negative impact on customers, of course we can include the titles and numbers.

kaitlynmichael:
> We're missing the following, is it on purpose?

Good catch! It is not on purpose. I'll add those asap.

amiramm:
> understand that we want to be consistent but I didn't think the Jira numbers and titles were adding any value on the public site. Internally, they are useful. For the public, I believe just a description of the problem is all that's necessary. If this has a significant negative impact on customers, we can include the titles and numbers.

@kaitlynmichael we use these ticket numbers with customers and they expect to see them. Otherwise, every time we would need to refer to an issue, we would need to repeat the entire description and that's just not practical. We need to keep those, please. Thanks.

amiramm (changes requested):
Thanks for this! please review comments/edits.

PR #1408, opened 2021-06-17
RC: Update references to old plan names. 

lanceleonard:
Primarily focused updates to plan names; also includes a few targeted copy edits.

We need to schedule a full dev edit/review on this content later.


lanceleonard:
These changes will be incorporated separately.

PR #1412, opened 2021-06-24
Update clustering.md

yanivhershkovich:
Removed the Redisearch limitation from number of shards section since this should be supported with RediSearch 2.
I'm not sure if we have other modules that currently don't support resharding. If we do, we should update the section to reflect it.


yanivhershkovich:
Another option:
As RediSearch 1.6.x is still out there, I don't know if we might want to keep the limitation and just mention the specific version. Something like (PMs to decide):

_An exception to this is a
RediSearch 1.6.x enabled database. Once you configure the number of
shards, you can no longer change that value. If you need to scale your
RediSearch 1.6.x enabled database, you must create a new database
at the new required size and replicate the current database to the new
one._

kaitlynmichael (approved):

PR #1416, opened 2021-06-28
update mDNS for rhel8

AlonMagrafta:
merge only when we add rhel8 support


kaitlynmichael (approved):

kaitlynmichael:
@AlonMagrafta  When are we adding rhel8 support?

AlonMagrafta:
working on it, should be by Aug 1st, I just added it now to not forget

PR #1417, opened 2021-06-28
Update crdb-cli-reference.md

yanivhershkovich:
The 'Adding an instance' section only includes a table of mandatory parameters but there are also optional parameters that can be added but are not mentioned.
For example: URL & replication_endpoint.

Here's a syntax example:
crdb-cli crdb add-instance --crdb-guid <GUID> --instance fqdn=<FQDN>,username=<USER>,password=<PASSWORD>,url=https://<FQDN>:9443,replication_endpoint=<FQDN>

Similar to the 'additional options' we provided in the 'crdb create' section, we should have something similar in the 'Adding an instance' section with an explanation of the optional parameters.

I only marked the place we should add the options in the page as I'm not aware of all the options.
Need R&D help here. Marked with:
============================================
The `crdb add-instance` command supports several additional options:
<ADD OPTIONS>
============================================


kaitlynmichael:
@yanivhershkovich since we don't know all the options we need to add yet, I've created a Jira and we can track this effort and collaborate with R&D from there (DOC-772)

PR #1449, opened 2021-07-14
added warning for persistent volume size

kaitlynmichael:
changes to warn customers about persistent volume size via DOC-767

### DIFF
--- a/content/platforms/kubernetes/concepts/persistent-volumes.md
+++ b/content/platforms/kubernetes/concepts/persistent-volumes.md
@@ -37,8 +37,12 @@ requirements]({{< relref "/rs/administering/designing-production/hardware-requir
 To explicitly specify the persistent storage size, use the *volumeSize*
 property as described in the example above.
 
+{{< warning >}}
+Be aware the persistent volume size cannot be changed after deployment. Trying to change this value after deployment can result in breaking the operator. Please be sure your specified *volumeSize* is correct at the time of creation.
+{{< /warning >}}
+
 {{< note >}}
-We recommed that you omit the volumeSize definition from the REC declaration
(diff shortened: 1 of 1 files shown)


rrelledge (approved):

content/platforms/kubernetes/concepts/persistent-volumes.md
```diff
@@ -37,8 +37,12 @@ requirements]({{< relref "/rs/administering/designing-production/hardware-requir
 To explicitly specify the persistent storage size, use the *volumeSize*
 property as described in the example above.
 
+{{< warning >}}
+Be aware the persistent volume size cannot be changed after deployment. Trying to change this value after deployment can result in breaking the operator. Please be sure your specified *volumeSize* is correct at the time of creation.
```
amiramm, line 41:
```suggestion
Be aware the persistent volume size cannot be changed after deployment. Trying to change this value after deployment can result in unexpected. Please be sure your specified *volumeSize* is correct at the time of creation.
```
Unexpected or undefined behavior rather than "break the operator".

amiramm (dismissed):
@kaitlynmichael I suggest you ask @yuvallevy2 to review as well

content/platforms/kubernetes/concepts/persistent-volumes.md
kaitlynmichael, line 41:
```suggestion
Be aware the persistent volume size cannot be changed after deployment. Trying to change this value after deployment could result in unexpected and potentially damaging behavior. Please be sure your specified *volumeSize* is correct at the time of creation.
```

yuvallevy2 (approved):

PR #1450, opened 2021-07-14
Fix image paths for "Encryption at rest" page

epicfaace:
Right now, the image paths on https://docs.redislabs.com/latest/rc/security/database-security/encryption-at-rest/ are broken:

image

Because they point to https://docs.redislabs.com/latest//images/rc/persistent-storage-encryption.png instead of https://docs.redislabs.com/latest/images/rc/persistent-storage-encryption.png.


lanceleonard (approved):
Looks good to me.

PR #1452, opened 2021-07-15
Updating RS version dropdown and redirects for RS 6.2.x release series

rrelledge:


kaitlynmichael (approved):
I'm definitely not very knowledgeable here but it looks good to me.

PR #1469, opened 2021-07-29
can add module only on creation

AlonMagrafta:


AlonMagrafta:
@K-Jo @emmanuelkeller @dragnot fyi

lanceleonard (approved):

PR #1478, opened 2021-08-08
Update _index.md

yorozen:
add a remark about RedisGraph does not support eviction policy.
Another note that can be added, is the best practice of using multiple modules in a single database.


kaitlynmichael:
I've created a Jira ticket to track this improvement internally (DOC-854).

PR #1487, opened 2021-08-13
Fixed incorrect GTM container and deduped the GA scripts

rrelledge:
Fixed incorrect GTM container and deduped the GA scripts


lanceleonard (approved):
LGTM

PR #1516, opened 2021-08-31
update broken links on K8s Openshift getting started

kaitlynmichael:


rrelledge (approved):
It looks like all of the broken links are fixed.

PR #1520, opened 2021-09-01
remove service broker ref, update version tag

amiramm:
Removed another reference to service broker spec.
Fix redisEnterpriseImageSpec reference to correct format, updated to latest versionTag (@kaitlynmichael I wonder if we can use a variable here to automatically retrieve the latest version number?)


kaitlynmichael:
> Removed another reference to service broker spec.
> Fix redisEnterpriseImageSpec reference to correct format, updated to latest versionTag (@kaitlynmichael I wonder if we can use a variable here to automatically retrieve the latest version number?)

I will work with Rachel to see if we have the infrastructure to use a variable for this.

kaitlynmichael (approved):

PR #1525, opened 2021-09-09
9081 is internal only

nail-sirazitdinov:
9081 is internal only. Active-Active management goes via reverse proxy (port 9443) and redirected to 9081.


lanceleonard (approved):
LGTM

PR #1537, opened 2021-09-15
Updated table-children shortcode to handle page description

rrelledge:
Example of shortcode with page description field:

{{< table-children columnNames="Name,Description" columnSources="LinkTitle,Description" enableLinks="LinkTitle" >}}


lanceleonard (approved):
LGTM

PR #1559, opened 2021-10-03
redis_upgrade_policy

gilbennun:
customer following these instructions will change the redis_upgrade_policy to latest.
this might be desirable for non-conservative customers, but not desirable for conservative customers.
we need to provide the customer with more details so that he can make the right decision.
Please add more details and make it clear that it is not mandatory to change policy to the latest.


kaitlynmichael:
I think the best place to track this work would be in a Jira ticket. I've created DOC-929 for this and added you as a watcher.

PR #1609, opened 2021-10-18
Update create-aa-database.md

yorozen:
Not clear,
If your cluster uses OpenShift routes, add the following to the `spec` section of your Redis Enterprise cluster (REC) resource file.

Need clarification if when using openshift we need to use openshiftRoute or we can choose between ingress and openshiftRoute


kaitlynmichael:
This was resolved by another PR.

PR #1634, opened 2021-11-01
Removed multi-key commands capability

guyco-redis:
Not used anywhere by the cluster or the proxy


lanceleonard:
@guyco-redis - This change was also requested in [PR #1635](https://github.com/RedisLabs/redislabs-docs/pull/1635), which has already been merged.

PR #1636, opened 2021-11-01
2+1nodes-cluster limitations

AlonMagrafta:


content/embeds/hardware-requirements-embed.md
```diff
@@ -23,15 +23,17 @@ We recommend these hardware requirements for production systems or for developme
 
 | Item | Description | Minimum Requirements | Recommended |
 |------------|-----------------|------------|-----------------|
-| Nodes per cluster | At least 3 nodes are required to support a reliable, highly available deployment that handles process failure, node failure, and network split events in a consistent manner. | 3 nodes | >= 3 nodes (Must be an odd number of nodes) |
+| Nodes per cluster<sup>*</sup> | At least 3 nodes are required to support a reliable, highly available deployment that handles process failure, node failure, and network split events in a consistent manner. | 3 nodes | >= 3 nodes (Must be an odd number of nodes) |
```
lanceleonard, line 26:
```suggestion
| Nodes per cluster<sup>*</sup> | At least three nodes are required to support a reliable, highly available deployment that handles process failure, node failure, and network split events in a consistent manner. | 3 nodes | >= 3 nodes (Must be an odd number of nodes) |
```

content/embeds/hardware-requirements-embed.md
```diff
@@ -23,15 +23,17 @@ We recommend these hardware requirements for production systems or for developme
...
 | Persistent Storage | Used for storing [snapshot (RDB format) and AOF files]({{< relref "/rs/administering/designing-production/persistent-ephemeral-storage.md" >}}) over a persistent storage media, such as AWS Elastic Block Storage (EBS) or Azure Data Disk. | RAM x 3 | In-memory >= RAM x 6 (except for [extreme 'write' scenarios]({{< relref "/rs/administering/designing-production/performance/disk-sizing-heavy-write-scenarios.md" >}})); [Redis on Flash]({{< relref "/rs/concepts/memory-architecture/redis-flash.md" >}}) >= (RAM + Flash) x 5. |
 | Network | We recommend using multiple NICs per node where each NIC is >100Kbps, but RS can also run over a single 1Gbps interface network used for processing application requests, inter-cluster communication, and storage access. | 1G | >=10G |
 
 <sup>*</sup>Additional considerations:
-
+- Nodes per Cluster:
+    - For clusters which are configured with 3 nodes, where one of them is considered as a quorum node - database availability can be compromised in a (non-quorum) node failure scenario.
```
lanceleonard, line 35:
```suggestion
    - When a cluster has three nodes and one of those node is considered a quorum node, database availability can be compromised when a node fails for non-quorum related reasons. 
```

content/embeds/hardware-requirements-embed.md
```diff
@@ -23,15 +23,17 @@ We recommend these hardware requirements for production systems or for developme
...
+    - Active-Active Redis databases are not recommended to run on a 3 node cluster, where one of them is considered as a quorum node. Synchronization process will pause in case of a (non-quorum) node failure scenario.
```
lanceleonard, line 36:
```suggestion
    - For best results, clusters with Active-Active Redis databases should have more than three nodes when one node is considered a quorum node. If a non-quorum node fails, there are no other nodes to sync to.  As a result, the synchronization process pauses until one becomes available.  
```

lanceleonard (changes requested):
I think we need a bit more detail.  How's this?

content/embeds/hardware-requirements-embed.md
AlonMagrafta: not exactly, it's when a node, which is not the quorum node, fails.
support call it "data-node" but I don't think that's an official term we use

content/embeds/hardware-requirements-embed.md
AlonMagrafta: It's not about the quorum node failing, it's about the other ones failing.
Basically we wanna say - for A/A better don't do a 3 node cluster with 1 of them functioning as quorum (aka 2 "data nodes" only)

tom-andrews: Why is that "not recommended"?
I think we should clearly state that with CRDB (Active-Active) in case the cluster only has 3 nodes, setting one of the nodes as quorum node is **not supported** since a single node failure can cause the synchronization process to be paused until the node becomes available.

content/embeds/hardware-requirements-embed.md
tom-andrews: Alon, IMHO "non-quorum node" should be fine in this case.
Something like:

For clusters consisting of only three nodes, with one of them configured as a quorum node - database redundancy can be compromised if a single non-quorum node fails.

I also switched "availability" to "redundancy", since the DB should be still accessible. Not sure whether we use this term in our docs, though.

lanceleonard (approved):
LGTM

PR #1663, opened 2021-11-16
Release note 2.2.5

emmanuelkeller:


emmanuelkeller:
Mistake. It should be set in a new file (major version = 2.2 not 2.0)

PR #1672, opened 2021-11-22
Removed old sitemap entry from robots.txt

rrelledge:


lanceleonard (commented):
LGTM (@banker, what do you think?)

PR #1681, opened 2021-11-29
Update billing-and-payments.md

yorozen:
Need to add the supported payment methods

do we support debit? prepaid? ...


lanceleonard:
@yorozen - Thanks for the suggestion; we'll take it under considerations.  In the future, please file suggestions via Jira.

PR #1684, opened 2021-11-30
Fixed JSON file  path name in curl invocation

monochromec:
Instead of incorrect "./static/code/rv/api/"


lanceleonard:
Hi @monochromec - Thanks for the contribution.  We need to review it carefully, as there's an internal process designed to validate these particular examples and we need to make sure that your change doesn't break it.  (The assumption is that developers would customize the examples to suit their own environments/requirements.)

We appreciate the help, though.  Thank you!  Feel free to  suggest other changes you think would help.

PR #1691, opened 2021-12-03
Allow toc parameter to force page TOC to appear

rrelledge:
You can now add `toc: "true"` to a page's front matter to force its TOC to appear.


lanceleonard (approved):
LGTM

PR #1693, opened 2021-12-06
Example for using a persistent volume claim

worldspawn:
I am of the opinion that showing how to to spin up redis insights in k8s with ephemeral storage only is not that useful. This is the solution I came up and is working well for me.

### DIFF
--- a/content/ri/installing/install-k8s.md
+++ b/content/ri/installing/install-k8s.md
@@ -181,14 +181,14 @@ spec:
           terminationMessagePolicy: File
       containers:
         - name:  redisinsight #Container name (DNS_LABEL, unique)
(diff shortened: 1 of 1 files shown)


lanceleonard:
@worldspawn - Thanks for the contribution; we'll check it out.

kaitlynmichael (approved):
I've made a few editorial suggested changes.

kaitlynmichael:
@worldspawn one of our support engineers gave me an alternate solution to the init container. He suggests that changint the security context for the pod would be cleaner. Would this solution work for you?
     ```
 securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        fsGroup: 2000
```

PR #1702, opened 2021-12-13
Revise RI K8s deployment article

kaitlynmichael:
Line editing and eliminating redundant code examples. Please review for technical accuracy and clarity. 
Jira: DOC-1040
Staging site: https://docs.redislabs.com/staging/jira-doc-1040/ri/installing/install-k8s/


content/ri/installing/install-k8s.md
```diff
@@ -9,210 +9,174 @@ nextStep:
...
+                chown -R 1001 /db
+            resources: {}
+            volumeMounts:
+              - name: db
+                mountPath: /db
+            terminationMessagePath: /dev/termination-log
+            terminationMessagePolicy: File
+      ```
+
+1. Add the following persistent volume claim to your deployment file (`redisinsight.yaml`).
```
bascar, line 158: May want to add something like "you may want to change the "default" storageClassName to your preferred storage class.

laurentdroin:
We should not explain how K8s Deployments work IMO. We can point at the K8s doc for this (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/).

Instead of 

```
The example below is a RedisInsight deployment file without a persistent volume claim or accompanying service. You’ll need to configure the following sections:
```

I would say something like:

```
The example below is a 1-replica RedisInsight deployment file without a persistent volume claim or accompanying service. See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/ for more information.
```

and that's it.

If we really want to explain some fields, I suppose we can but we should really not discuss things like:

```
..selector: The selector label (matchLabels.app) determines which pods the deployment will manage. This must to match the pod template label (spec.template.metadata.labels.app`).
```

or

```
...template.spec.containers: Set the container name, image, and image pull policy for the pod template. The container name must be unique and compatible with your DNS labels.
```

Someone who wants to deploy in K8s must know the basics, and these are the basics. 
We provide a sample YAML, and if they want to understand why it is the way it is, they can check the k8s doc.


That's only my opinion of course.

laurentdroin:
For the service, I see that we are suggesting a LoadBalancer service and that's ok. We might want to indicate it may take a few minutes for the Loadbalancer to be provisioned and the DNS to be updated.
Also, we could indicate how to reach RedisInsight: `http://<external-ip>`

Note that the external IP can be a hostname:

```
NAME                   TYPE           CLUSTER-IP      EXTERNAL-IP                                                               PORT(S)        AGE
redisinsight-service   LoadBalancer   10.43.106.224   ad01707e415fa4c2683e6f60171407a6-1783640786.us-east-1.elb.amazonaws.com   80:32131/TCP   6m28s
```

content/ri/installing/install-k8s.md
laurentdroin: +1 - the storage class name is key here and has to be a conscious choice from the user. Plus there might not be a default storage class.

laurentdroin:
In "Create a RedisInsight service "

if the block of yaml is meant to be appended to the main yaml file, then let's insert "---" as the first line, just like there is for the PVC. That's how you separate multiple k8s resource definitions in the yaml file. Customers hopefully know this but it's good to be consistent.

laurentdroin:
I think this :

```
a. Set the replica strategy (spec.replicas.strategy) Recreate. This will avoid multiple pods trying to use the same volume.

```
is not correct.
What we need to change is .spec.strategy.type

The small yaml seems more correct but it's out of alignment and is out of context (i.e. you don't really see where it fits there. It should be 1 indentation level below "spec".

laurentdroin:
Instead of the init container (which feels a bit hacky), I wonder if setting a security context on the pod would not be cleaner.

I tested with this on my Rancher cluster and it worked fine:

```
      securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        fsGroup: 2000
```

More info at https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
It could be worth reaching out to the PR submitter and asking them whether this would work for them.

kaitlynmichael:
@laurentdroin and @bascar thank you for the feedback. I've made all the changes you suggested.
You can view the preview here: https://docs.redislabs.com/staging/jira-doc-1040/ri/installing/install-k8s/

content/ri/installing/install-k8s.md
```diff
@@ -9,210 +9,203 @@ nextStep:
...
 
-## Create the RedisInsight deployment without a service.
+Make sure to substitute your own values into your deployment file. Go to [kubernetes.io](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) for more information on deployments.
 
-Below is an annotated YAML file that will create a RedisInsight
-deployment in a K8s cluster.
-
-1. Create a new file redisinsight.yaml with the content below
+{{<note>}}If you have a service named `redisinsight` that exposes the deployment, it may cause environment variable conflicts with the RedisInsight software. You can manually override the service environnement variables (as seen in this example) or change the service name (as seen in the service example below).{{</note>}}
```
rrelledge, line 23:
```suggestion
{{<note>}}If you have a service named `redisinsight` that exposes the deployment, it may cause environment variable conflicts with the RedisInsight software. You can manually override the service environment variables (as seen in this example) or change the service name (as seen in the service example below).{{</note>}}
```

content/ri/installing/install-k8s.md
```diff
@@ -9,210 +9,203 @@ nextStep:
...
+1. Make the following updates to your deployment spec:
+
+    a. Set the strategy (`spec.strategy.type`) to [`Recreate`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment). This will avoid multiple pods trying to use the same volume.
+
+      ```yaml
+        strategy:
+          type: Recreate
+      ```
+  
+    b. In the `.template.spec.volumes` section, replace the `emptyDir` volume the name of your persistent volume claim.
```
rrelledge, line 115:
```suggestion
    b. In the `.template.spec.volumes` section, replace the `emptyDir` volume with the name of your persistent volume claim.
```

PR #1729, opened 2022-01-05
Support offline install steps - updated

shahar-biron:
According to the latest 6.2 offline install options.
@MeirShpilraien please review

### DIFF
--- a/content/modules/redisgears/installing-redisgears.md
+++ b/content/modules/redisgears/installing-redisgears.md
@@ -17,11 +17,11 @@ Before you can use RedisGears, you have to install the RedisGears module on your
 ### Step 1: Install RedisGears dependencies
 
 {{< note >}}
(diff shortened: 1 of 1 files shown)


MeirShpilraien (commented):
This is not a correct fix, the instruction for Redis 6.0.12 and above are on the link specified and it also contains a note what to do on offline environment.

K-Jo:
Let's not close this until @MeirShpilraien approved please.

MeirShpilraien:
@K-Jo as I commented above, this change is wrong and misleading. Offline installation on 6.0.12 and above are describe here: https://docs.redis.com/latest/modules/add-module-to-cluster/. If something is not clear we should update this page.

lanceleonard:
Closing this PR in favor of further review.

PR #1755, opened 2022-01-18
Different output of slowlog get for Redis OSS

igordr3:
SLOWLOG GET command features different output for Redis OSS compared to Redis Enterprise/Cloud.

In OSS Redis, the slowlog get command gives the following fields:

1. A unique progressive identifier for every slow log entry.
2. The unix timestamp at which the logged command was processed.
3. The amount of time needed for its execution, in microseconds.
4. The array composing the arguments of the command.
5. Client IP address and port.
6. Client name if set via the CLIENT SETNAME command.

Whereas in our Redis Enterprise Cloud DB, the slowlog get command gives the following fields:

1. A unique progressive identifier for every slow log entry.
2. The unix timestamp at which the logged command was processed.
3. The amount of time needed for its execution, in microseconds.
4. The array composing the arguments of the command.

Hence the documentation to reflect it till we decide on how to proceed further.

### DIFF
--- a/content/rc/faqs/_index.md
+++ b/content/rc/faqs/_index.md
@@ -80,6 +80,8 @@ Every effort is made to adhere, where possible, to the specifications of open so
 
 - Open source Redis supports key names up to 512MB.  Redis Enterprise Software and Redis Enterprise Cloud each limit key names to 64KB.  Key values can reach up to 512MB in all three services.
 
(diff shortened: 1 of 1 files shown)


igordr3:
@lanceleonard Could you review when convenient, please?

lanceleonard:
@igordr3 - I took a quick look when i assigned the PR to me.  Thanks for the contribution; we appreciate the help!

I've already reached out to our product team to verify the info and how we should present it.  (We've been trying to think about a better way to present command compatibility info and the FAQs don't really feel like the right place for this level of detail.)

lanceleonard:
@AlonMagrafta - Please take a look.

AlonMagrafta (changes requested):
wow missed that one for a long time
I suggest to add examples of the output

PR #1775, opened 2022-02-04
Added offline installation for RedisGears JVM

rrelledge:
[Staged preview](https://docs.redis.com/staging/jira-doc-1139/modules/redisgears/jvm/install/#install-redisgears-and-the-jvm-plugin)

### DIFF
--- a/content/modules/redisgears/jvm/install.md
+++ b/content/modules/redisgears/jvm/install.md
@@ -20,19 +20,36 @@ Before you can use RedisGears with the JVM, you need to install the RedisGears m
 
 ## Install RedisGears and the JVM plugin
 
-If your cluster uses Redis Enterprise v6.0.12 or later, you only need to download the RedisGears package. The JVM plugin installs automatically when you make a `POST` request to the `/v2/modules` REST API endpoint.
+If your cluster uses Redis Enterprise v6.0.12 or later and has internet access, you only need to download the RedisGears package. It automatically fetches dependencies like the JVM plugin during online installation.
 
-After the install is complete, RedisGears will appear in the list of available modules on the **settings** and **create database** pages of the Redis Enterprise admin console.
+Offline installation requires you to manually upload dependencies to the master node.
+
+To install RedisGears and the JVM plugin:
 
 1. Download the RedisGears package from the [download center](https://redis.com/redis-enterprise-software/download-center/modules/).
 
-1. Upload the package to a node in the cluster.
+    {{<note>}}
+For offline installation, you also need to download the packages for RedisGears dependencies.
+    {{</note>}}
 
-1. Add RedisGears to the cluster with a `POST` request to the `/v2/modules` REST API endpoint:
+1. Upload the RedisGears package to a node in the cluster.
+
+1. For offline installation only, copy the dependencies to the following directory on the master node: `$modulesdatadir/rg/<version_integer>/deps/`
+    ```sh
+    $ cp gears-jvm.linux-centos7-x64.0.1.0.tgz $modulesdatadir/rg/10201/deps/
+    ```
+
+    {{<note>}}
+Skip this step unless your cluster does not have internet access. 
+    {{</note>}}
+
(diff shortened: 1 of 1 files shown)


content/modules/redisgears/jvm/install.md
```diff
@@ -20,20 +20,37 @@ Before you can use RedisGears with the JVM, you need to install the RedisGears m
...
-After the install is complete, RedisGears will appear in the list of available modules on the **settings** and **create database** pages of the Redis Enterprise admin console.
+Offline installation requires you to manually upload dependencies to the master node.
+
+To install RedisGears and the JVM plugin:
 
 1. Download the RedisGears package from the [download center](https://redis.com/redis-enterprise-software/download-center/modules/).
 
-1. Upload the package to a node in the cluster.
+    {{<note>}}
+For offline installation, you also need to download the packages for RedisGears dependencies.
```
MeirShpilraien, line 32: I do not know how it looks on the download page, maybe we should also state the files named (if its clear on the download page then its not needed)...

MeirShpilraien (approved):
Small suggestion, other then this looks good :+1:

content/modules/redisgears/installing-redisgears.md
```diff
@@ -9,21 +9,18 @@ Before you can use RedisGears, you have to install the RedisGears module on your
...
-{{< note >}}
-This step is required for Redis Enterprise Software 6.0.8 and below only.
-For RS 6.0.12 and above, the dependencies are installed automatically when [you install with the `/v2/modules` endpoint]({{< relref "/modules/add-module-to-cluster.md#adding-a-module-using-the-rest-api" >}}).
-{{< /note >}}
+This step is optional and is only required for environments that don't have direct access to the internet for downloading the dependencies.
 
 On each node in the Redis Enterprise cluster:
 
-1. [Download](https://redislabs.com/download-center/modules/) the RedisGears Module - Dependencies Package from the Redis Enterprise Software section of the Downloads page.
+1. [Download](https://redislabs.com/download-center/modules/) the RedisGears Module - Dependencies Package for both Java and Python from the Redis Enterprise Software section of the Downloads page.
```
MeirShpilraien, line 23: If we are talking about 6.0.12 this is no longer true ...

content/modules/redisgears/jvm/install.md
```diff
@@ -14,25 +14,12 @@ Before you can use RedisGears with the JVM, you need to install the RedisGears m
...
-1. Created a Redis Enterprise cluster
+1. Created a Redis Enterprise cluster TODO link
 
-1. Added nodes to the cluster
+1. Added nodes to the cluster TODO link
 
-## Install RedisGears and the JVM plugin
+1. Install RedisGears and the JVM plugin TODO link to installation instructions
 
-If your cluster uses Redis Enterprise v6.0.12 or later, you only need to download the RedisGears package. The JVM plugin installs automatically when you make a `POST` request to the `/v2/modules` REST API endpoint.
```
MeirShpilraien, line 23: the change @rrelledge did on this PR (before @K-Jo changes) are correct, we should just move them to installing-redisgears.md page (I guess...)

K-Jo: Yup, I see I forgot to move this

rrelledge: I moved this info to the requested page.

lanceleonard (approved):
LGTM

K-Jo (changes requested):
LGTM but, I would still move the section of "write behind" at the top level to a "recipe" section under python.  Right now it looks at the top level that it can't be done with jvm.  While there is a dedicated recipe section under JVM.  The only thing we need to duplicate between two sections of write behind (jvm and python) is the install of the oracle driver if I'm not mistaken. @MeirShpilraien 

so two things
- installing-redisgeaers.md at top level should no longer talk about the write-behind. This part should go in write-behind.md at that level
- this write-behind.md level should become a recipe file like it is in jvm but then under python

PR #1793, opened 2022-02-25
The cluster license is not stored in a secret

laurentdroin:


laurentdroin:
Ah never mind, it is still possible to create a secret to hold the license and refer to this secret so closing that PR. No change needed.

PR #1830, opened 2022-03-27
Update cluster-options.md

erniavr:
One of our customers encountered an error when trying to use floats as sizing numbers.
added the note above to avoid future problems with other customers.


kaitlynmichael (approved):
Thanks for adding this!

PR #1833, opened 2022-03-28
DOC-1239 integrate cluster option notes

kaitlynmichael:
Integrated redisEnterpriseNodeResources notes into main text. I also clarified that both limits and requests are integer values.


lanceleonard (approved):
LGTM

PR #1847, opened 2022-04-01
DOC-1132: RS - Examples of "crdb create examples" using --encrypted rather than --encryption

cmilesb:
Staging Link: https://docs.redis.com/staging/jira-doc-1132/rs/references/crdb-cli-reference/

- Fixed code samples to use the "--encryption" argument instead of "--encrypted".
- I noticed that the table under "Updating the configuration of an Active-Active database" had some pipes that weren't escaped, which caused some formatting issues on the latest version of the site (see screenshot for current latest look). Escaped those pipes so that the table looks better.
Screen Shot 2022-04-01 at 5 02 16 PM
- In that same table, the "--encryption" argument had the "binary" type when I assume it should be "boolean" (like the other true/false arguments).


rrelledge (approved):
Nice catches! Looks good to me.

rrelledge (approved):
Looks good to me.

PR #1881, opened 2022-04-19
Update usage-reports.md

AnnaRedis:
Can you please update based on the new UI and the fact that there are 2 reports, one is for memory usage and the other one is based on the shard usage.


lanceleonard:
Closing, as this work was included in [PR #2010](https://github.com/RedisLabs/redislabs-docs/pull/2010).

PR #1922, opened 2022-05-12
Reorg rladmin cluster commands

rrelledge:
[Staged preview](https://docs.redis.com/staging/jira-doc-1316/rs/references/cli-utilities/rladmin/cluster/)

This PR also includes an update to the `table-children` shortcode. This change allows you to limit which child pages appear in a generated table.

For example, I added either `tags: ["configured"]` or `tags: ["non-configured"]` to the front matter of each `rladmin cluster` command page.

To generate a table that only includes commands that are tagged as "configured," I used this shortcode on the `rladmin cluster` index page and set the new optional field `limitTags` to "configured":

```sh
{{<table-children columnNames="Command,Description" columnSources="linkTitle,Description" enableLinks="linkTitle" limitTags="configured">}}
```

### DIFF
--- a/content/rs/references/cli-utilities/rladmin/cluster/_index.md
+++ b/content/rs/references/cli-utilities/rladmin/cluster/_index.md
@@ -10,23 +10,12 @@ categories: ["RS"]
 aliases: 
 ---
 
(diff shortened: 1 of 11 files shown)


cmilesb (changes requested):
Found a couple of things that might need to be changed, but otherwise looks good!

cmilesb (approved):
Looks good to me!

PR #1928, opened 2022-05-18
Changes in Ingress spec in Kubernetes 1.19

NLxAROSA:
pathType is now mandatory; provided a default of Prefix.
serviceName has been replaced by service.name
servicePort has been replaced by service.port.number

### DIFF
--- a/content/kubernetes/re-databases/set-up-ingress-controller.md
+++ b/content/kubernetes/re-databases/set-up-ingress-controller.md
@@ -88,12 +88,12 @@ Install one of the supported ingress controllers:
         http:
           paths:
           - path: /
-            pathType: Prefix
+            pathType: ImplementationSpecific
             backend:
               service:
                 name: <db-name>
                 port:
-                  number: <db-port>
+                  name: redis
     ```  
 
     For HAProxy, insert the following into the `annotations` section:  
--- a/content/kubernetes/re-databases/set-up-ingress-controller.md
+++ b/content/kubernetes/re-databases/set-up-ingress-controller.md
@@ -88,9 +88,12 @@ Install one of the supported ingress controllers:
         http:
           paths:
           - path: /
+            pathType: Prefix
             backend:
-              serviceName: <db-name>
-              servicePort: <db-port>
+              service:
+                name: <db-name>
+                port:
+                  number: <db-port>
     ```  
 
     For HAProxy, insert the following into the `annotations` section:  


content/kubernetes/re-databases/set-up-ingress-controller.md
```diff
@@ -88,9 +88,12 @@ Install one of the supported ingress controllers:
...
           paths:
           - path: /
+            pathType: Prefix
             backend:
-              serviceName: <db-name>
-              servicePort: <db-port>
+              service:
+                name: <db-name>
+                port:
+                  number: <db-port>
```
zcahana, line 96: `port.number` is valid, but we could also instruct users to set `port.name: redis` instead, which is the default port name used in the REDB service. This should work out of the box and spare the user from disovering the REDB service port.

content/kubernetes/re-databases/set-up-ingress-controller.md
```diff
@@ -88,9 +88,12 @@ Install one of the supported ingress controllers:
         http:
           paths:
           - path: /
+            pathType: Prefix
```
zcahana, line 91: On one hand, the operator is creating the ActiveActive ingresses with `pathType: ImplementationSpecific` so it would be nice to be on par with it. And anyway's Nginx's behavior for `ImplementationSpecific` is identical to `Prefix` and HAProxy's behavior is quite close too (`path-type: begin`).

On the other hand, `Prefix` is somewhat better match semantically, especially if this inspires the user how to create similar ingress resources for the cluster API/UI which are actually HTTP.

On yet another hand, this service is TCP-based so this hardly even matters.

/shrug

zcahana: Bottom line, I suggest to stick to `pathType: ImplementationSpecific`.

content/kubernetes/re-databases/set-up-ingress-controller.md
```diff
@@ -88,9 +88,12 @@ Install one of the supported ingress controllers:
...
             backend:
-              serviceName: <db-name>
-              servicePort: <db-port>
+              service:
+                name: <db-name>
+                port:
+                  number: <db-port>
     ```  
 
     For HAProxy, insert the following into the `annotations` section:
```
zcahana, line 99: Here below, we instruct the user to set the `kubernetes.io/ingress.class` annotation, which too is deprecated since 1.19 or so.  Instead, the user should set the `.spec.ingressClassName` field with a similar `haproxy`/`nginx` value.

content/kubernetes/re-databases/set-up-ingress-controller.md
NLxAROSA: I will change this to ImplementationSpecific.

content/kubernetes/re-databases/set-up-ingress-controller.md
zcahana: Testing is always a good idea, especially when feeling (too) comfortable.

kaitlynmichael:
@NLxAROSA have you completed your tests to verify these changes yet?

NLxAROSA:
> @NLxAROSA have you completed your tests to verify these changes yet?

Yes, tested and verified. Added commit to reflect the changes.

kaitlynmichael (approved):
Thank you @NLxAROSA and @zcahana for your help on this!

PR #1931, opened 2022-05-20
Update memory-optimizations.md

richzw:
Try to remove the incorrect fourth method of Reclaim expired keys memory faster

### DIFF
--- a/content/ri/memory-optimizations.md
+++ b/content/ri/memory-optimizations.md
@@ -69,7 +69,6 @@ You can follow one of these three steps to reclaim the memory:
 1. Restart your redis-server
 1. Increase memorysamples in redis conf. (default is 5, max is 10) so that expired keys are reclaimed faster.
 1. You can set up a cron job that runs the scan command after an interval which helps in reclaiming the memory of the expired keys.
-1. Alternatively, Increasing the expiry of keys also helps.
 
 #### Trade offs
 
--- a/content/ri/memory-optimizations.md
+++ b/content/ri/memory-optimizations.md
@@ -67,7 +67,7 @@ Looking at the difference you can clearly point out that lot of memory is still 
 You can follow one of these three steps to reclaim the memory:
 
 1. Restart your redis-server
-1. Increase `maxmemory-samples` in redis.conf (default is 5, max is 10) so that the expired keys are reclaimed faster.
+1. To reclaim the expired keys faster, increase memory-samples in redis.conf (default is 5).
 1. You can set up a cron job that runs the scan command after an interval which helps in reclaiming the memory of the expired keys.
 
 #### Trade offs
--- a/content/ri/memory-optimizations.md
+++ b/content/ri/memory-optimizations.md
@@ -72,7 +72,7 @@ You can follow one of these three steps to reclaim the memory:
 
 #### Trade offs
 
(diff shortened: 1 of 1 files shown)


content/ri/memory-optimizations.md
```diff
@@ -69,7 +69,6 @@ You can follow one of these three steps to reclaim the memory:
 1. Restart your redis-server
 1. Increase memorysamples in redis conf. (default is 5, max is 10) so that expired keys are reclaimed faster.
```
sundb, line 70: Maybe we can change `memorysamples` to `maxmemory-samples` by the way.
There is another one below.

```suggestion
1. Increase maxmemory-samples in redis conf. (default is 5, max is 10) so that expired keys are reclaimed faster.
```

nermiller, line 70:
```suggestion
1. Increase 'maxmemory-samples` in redis.conf (default is 5, max is 10) so that the expired keys are reclaimed faster.
```

nermiller (commented):
Hi, @sundb, please review my comment and let me know your thoughts. Also, please sign our CLA when you get the chance. Thanks for contributing! - Nermina

content/ri/memory-optimizations.md
sundb: '(here should be \` )maxmemory-samples`.

nermiller: You're right. Thanks!

sundb: `max is 10` is wrong, 10 only means better accuracy, but will consume more cpu, its max value is int_max.

content/ri/memory-optimizations.md
```diff
@@ -67,7 +67,7 @@ Looking at the difference you can clearly point out that lot of memory is still
 You can follow one of these three steps to reclaim the memory:
 
 1. Restart your redis-server
-1. Increase memorysamples in redis conf. (default is 5, max is 10) so that expired keys are reclaimed faster.
+1. Increase `maxmemory-samples` in redis.conf (default is 5, max is 10) so that the expired keys are reclaimed faster.
```
sundb, line 70: `max is 10` is wrong, 10 only means better accuracy, but will consume more cpu, its max value is int_max.

sundb: Here is another `memorysamples`.
```
If we increase the memorysamples config, it expires the keys faster, but it costs more CPU cycles, which increases latency of commands. Secondly, increasing the expiry of keys helps but that requires significant changes to application logic.
```

nermiller (approved):
LGTM!

content/ri/memory-optimizations.md
nermiller: @sundb, should we say "To reclaim the expired keys faster, increase `memory-samples` in redis.conf (default is 5)."

nermiller (commented):
left a comment above

content/ri/memory-optimizations.md
sundb: seems good to me.

nermiller (approved):
Revised the sentence and updated the other mention of maxmemory-samples

nermiller:
@sundb please give me another check. Thanks!

sundb:
LGTM

PR #2016, opened 2022-06-30
K8s 6.2.10-45 Lucy release notes and distro changes 

kaitlynmichael:
Jira: DOC-1498
Staging pages:

-  https://docs.redis.com/staging/jira-doc-1498/kubernetes/release-notes/k8s-6-2-10-45-2022-07/
- https://docs.redis.com/staging/jira-doc-1498/kubernetes/reference/supported_k8s_distributions/

### DIFF
--- a/content/kubernetes/reference/supported_k8s_distributions.md
+++ b/content/kubernetes/reference/supported_k8s_distributions.md
@@ -36,16 +36,16 @@ Each release of the Redis Enterprise operator is thoroughly tested against a set
 - Any distribution not listed below is not supported for production workloads.
 
 
-| **Kubernetes version**  | 1.19       | 1.20       | 1.21       | 1.22       | 1.23       |
-|:------------------------|:----------:|:----------:|:----------:|:----------:|:----------:|
-| Community Kubernetes    | deprecated | deprecated | supported  | supported  | supported* |
-| Amazon EKS              | supported  | supported  | supported  |            |            |
-| Azure AKS               |            |            | supported  | supported  | supported* |
-| Google GKE              | supported  | supported  | supported  | supported  |            |
-| Rancher 2.6             | supported  | supported  | supported  | supported  |            |
+| **Kubernetes version**  | 1.19       | 1.20       | 1.21       | 1.22       | 1.23       | 1.24       |
+|:------------------------|:----------:|:----------:|:----------:|:----------:|:----------:|:----------:|
+| Community Kubernetes    |            |            | deprecated | supported  | supported  | supported* |
+| Amazon EKS              | deprecated | deprecated | supported  | supported* |            |
+| Azure AKS               |            |            | deprecated | supported  | supported  |
+| Google GKE              | deprecated | deprecated | supported  | supported  |supported*  |
+| Rancher 2.6             | deprecated | deprecated | supported  | supported  |            |
 | **OpenShift version**   | **4.6**    | **4.7**    | **4.8**    | **4.9**    | **4.10**   |
-|                         | deprecated | deprecated | supported  | supported  | supported* |
+|                         |            | deprecated | deprecated | supported  | supported  |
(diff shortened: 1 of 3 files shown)


content/kubernetes/release-notes/k8s-6-2-10-45-2022-07.md
```diff
@@ -0,0 +1,151 @@
...
+
+The Redis Enterprise K8s 6.2.10-45 supports the Redis Enterprise Software release 6.2.10 and includes feature improvements and bug fixes.
+
+The key bug fixes and known limitations are described below.
+
+## Images
+
+This release includes the following container images:
+
+* **Redis Enterprise**: `redislabs/redis:6.2.10-129` or  `redislabs/redis:6.2.10-129.rhel8-openshift`
```
zcahana, line 22: We publish both `redislabs/redis:6.2.10-129.rhel8-openshift` and `redislabs/redis:6.2.10-129.rhel7-openshift` images for this release (also for Kraken 6.2.10-34 with RS 6.2.10-107).

I believe the instructions to users should be to stick to RHEL7 when upgrading from a RHEL7 cluster, and use RHEL8 for new clusters as well as upgrades from RHEL8. Maybe this can go into the "known limitations" in the Lucy release notes since we do have validation and enforcement that no RHEL7-->RHEL8 upgrade is attempted. But in the Kraken release notes this should remain as a warning since we have no such validation there.

zcahana: @bascar @yuvallevy2 what do you think about that?

kaitlynmichael, line 22:
```suggestion
* **Redis Enterprise**: `redislabs/redis:6.2.10-129` or  `redislabs/redis:6.2.10-129.rhel7-openshift`
```

kaitlynmichael: @zcahana should we only mention the rhel7 one to avoid confusion?

zcahana: I think we should try to find the least confusing way to mention both, since rhel8 is the primary image and is the desired one looking forward, and rhel7 is only there as a workaround, hopefully temporary.

content/kubernetes/reference/supported_k8s_distributions.md
```diff
@@ -36,16 +36,16 @@ Each release of the Redis Enterprise operator is thoroughly tested against a set
...
-| **Kubernetes version**  | 1.19       | 1.20       | 1.21       | 1.22       | 1.23       |
-|:------------------------|:----------:|:----------:|:----------:|:----------:|:----------:|
-| Community Kubernetes    | deprecated | deprecated | supported  | supported  | supported* |
-| Amazon EKS              | supported  | supported  | supported  |            |            |
-| Azure AKS               |            |            | supported  | supported  | supported* |
-| Google GKE              | supported  | supported  | supported  | supported  |            |
-| Rancher 2.6             | supported  | supported  | supported  | supported  |            |
+| **Kubernetes version**  | 1.19       | 1.20       | 1.21       | 1.22       | 1.23       | 1.24       |
+|:------------------------|:----------:|:----------:|:----------:|:----------:|:----------:|:----------:|
+| Community Kubernetes    |            |            | deprecated | supported  | supported  | supported* |
```
zcahana, line 41: Vanilla Kubernetes 1.20 should be reinstated here (as deprecated), since in the previous release it was inconsistently mentioned as supported (in k8s-docs) and deprecated (official docs). Let's take the conservative approach and have it deprecated for yet another round before being removed. In terms of actual support - it was tested and certified throughout this release similarly to the rest of the supported versions, so we're good with that.

PR #2017, opened 2022-07-05
Update upgrading.md

yorozen:
Due to permissions issue need to run as : sudo $bindir/rlcheck


AlonMagrafta:
@yorozen is it always? (I didn't check but wanted to double verify with you as you are dealing with tons of these)

rrelledge:
This issue was fixed in the recent [6.4.2-81 release](https://docs.redis.com/latest/rs/release-notes/rs-6-4-2-releases/rs-6-4-2-81/#resolved-issues). `rlcheck` should now work without using `sudo`.

PR #2022, opened 2022-07-06
K8s: OpenShift routes article DOC-883

kaitlynmichael:
Staging: https://docs.redis.com/staging/jira-doc-883/kubernetes/re-databases/routes/
Jira: DOC-883

### DIFF
--- a/content/kubernetes/re-databases/routes.md
+++ b/content/kubernetes/re-databases/routes.md
@@ -11,9 +11,9 @@ aliases: [
 ]
 ---
 
-Every time a Redis Enterprise database (REDB) is created with the Redis Enterprise operator, a [service](https://kubernetes.io/docs/concepts/services-networking/service/) is created that allows requests to be routed to that database. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
+Every time the Redis Enterprise operator creates a Redis Enterprise database (REDB), it creates a [service](https://kubernetes.io/docs/concepts/services-networking/service/) that allows requests to be routed to that database. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
 
-By default, REDB creates a `ClusterIP` type service, which exposes a cluster-internal IP and can only be accessed from within the cluster. OpenShift routes allow requests to be routed to the REDB from outside the cluster. For OpenShift deployments, routes provides a preferred alternative to an ingress.
+By default, REDB creates a `ClusterIP` type service, which exposes a cluster-internal IP and can only be accessed from within the cluster. OpenShift routes allow requests to be routed to the REDB from outside the cluster. For OpenShift deployments, routes provide a preferred alternative to an ingress.
 
 ## Prerequisites
 
@@ -37,15 +37,15 @@ By default, REDB creates a `ClusterIP` type service, which exposes a cluster-int
   - **TLS Termination**: Choose "passthrough"
   - **Insecure Traffic**: Select "None"
 
-  Select **Create**.
+1. Select **Create**.
 
-1. Find the hostname for your new route. After route creation, it will be in the "Host" field.
+1. Find the hostname for your new route. After route creation, it appears in the "Host" field.
 
 1. Verify you have a DNS entry to resolve the hostname for your new route to the cluster's load balancer.
 
 ## Access database
 
-1. Use the `openssl` command to access the database from outside the cluster.
+Use the `openssl` command to access the database from outside the cluster.
 
   ```sh
   openssl s_client -connect <hostname>:443 -crlf -CAfile ./ca.pem -servername <hostname>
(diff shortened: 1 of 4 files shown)


laurentdroin:
In "routes provides a preferred alternative to ingress controller." I would remove "controller". A route is an equivalent to an ingress. Both routes and ingress are k8s resources and are very similar.
The equivalent to an ingress controller would be the (built-in) openshift router.


"Create a DNS entry to resolve the hostname for your new route to the cluster’s load balancer." This should be optional so I would not ask customers to do this. It should normally be done using a generic entry created at openshift installation time. That's why in my initial notes, I wrote "make sure" (i.e. verify, not create). 


"Make sure your database had TLS encryption enabled." This should not be in the part where you access the database but is a pre-requisite. If you want to access your database from outside k8s, whether using a Route or an Ingress, the database must be TLS enabled. So maybe we need to have this at the top of the page. Just like we have in the doc about "Ingress".

kaitlynmichael:
@laurentdroin I've made the changes you suggested and this is ready for another review.

content/kubernetes/re-databases/routes.md
```diff
@@ -0,0 +1,54 @@
...
+categories: ["Platforms"]
+aliases: [
+   /kubernetes/re-databases/routes.md,
+   /kubernetes/re-databases/routes/,
+]
+---
+
+Every time a Redis Enterprise database (REDB) is created with the Redis Enterprise operator, a [service](https://kubernetes.io/docs/concepts/services-networking/service/) is created that allows requests to be routed to that database. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
+
+By default, REDB creates a `ClusterIP` type service, which exposes a cluster-internal IP and can only be accessed from within the cluster. OpenShift routes allow requests to be routed to the REDB from outside the cluster. For OpenShift deployments, routes provides a preferred alternative to an ingress.
```
rrelledge, line 16:
```suggestion
By default, REDB creates a `ClusterIP` type service, which exposes a cluster-internal IP and can only be accessed from within the cluster. OpenShift routes allow requests to be routed to the REDB from outside the cluster. For OpenShift deployments, routes provide a preferred alternative to an ingress.
```

content/kubernetes/re-databases/routes.md
```diff
@@ -0,0 +1,54 @@
...
+  - **Name**: Choose any name you want as the first part of your generated hostname
+  - **Hostname**: Leave blank
+  - **Path**: Leave as is ("/")
+  - **Service**: Select the service for the database you want to access
+  - **TLS Termination**: Choose "passthrough"
+  - **Insecure Traffic**: Select "None"
+
+  Select **Create**.
+
+1. Find the hostname for your new route. After route creation, it will be in the "Host" field.
```
rrelledge, line 42:
```suggestion
1. Find the hostname for your new route. After route creation, it appears in the "Host" field.
```

content/kubernetes/re-databases/routes.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. Verify you have a DNS entry to resolve the hostname for your new route to the cluster's load balancer.
+
+## Access database
+
+1. Use the `openssl` command to access the database from outside the cluster.
```
rrelledge, line 48: Maybe remove the numbered list since it looks like there's only one step in this section.
```suggestion
Use the `openssl` command to access the database from outside the cluster.
```

content/kubernetes/re-databases/routes.md
```diff
@@ -0,0 +1,54 @@
...
+weight: 15
+alwaysOpen: false
+categories: ["Platforms"]
+aliases: [
+   /kubernetes/re-databases/routes.md,
+   /kubernetes/re-databases/routes/,
+]
+---
+
+Every time a Redis Enterprise database (REDB) is created with the Redis Enterprise operator, a [service](https://kubernetes.io/docs/concepts/services-networking/service/) is created that allows requests to be routed to that database. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
```
rrelledge, line 14: Does the operator create the database? Does the operator also create the service? If so, maybe reword this slightly?
```suggestion
Every time the Redis Enterprise operator creates a Redis Enterprise database (REDB), it creates a [service](https://kubernetes.io/docs/concepts/services-networking/service/) that allows requests to be routed to that database. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
```

nermiller, line 14:
```suggestion
When you create a Redis Enterprise database (REDB) using the Redis Enterprise operator, a [service](https://kubernetes.io/docs/concepts/services-networking/service/) that allows requests to be routed to that database is generated. Redis Enterprise supports three [types of services](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) for accessing databases: `ClusterIP`, `headless`, or `LoadBalancer`.
```

content/kubernetes/re-databases/routes.md
nermiller: Consider also adding Red Hat before _OpenShift_.

nermiller (approved):
Just a couple of editorial nits.

PR #2042, opened 2022-07-21
Added ranges to RC manage subscriptions table

Sawdustie:
https://redislabs.atlassian.net/browse/DOC-1532

Updated manage subscriptions table to include ranges, this should more accurately reflect the differences between each subscription plan.


cmilesb (approved):
Looks good!

PR #2058, opened 2022-07-27
Fixed links on Prometheus integration

Sawdustie:
Fixed links so that the url is not visible on the page


rrelledge (approved):
Looks good to me.

PR #2061, opened 2022-07-29
Updated RedisJSON quick start

rrelledge:
Updated RedisJSON quick start:
- Updated examples to use JSONPath instead of legacy path
- Streamlined/clarified examples
- Fixed reported bugs (DOC-1515)
- Copy edits
- Added links for commands and JSON path syntax

[Staged preview](https://docs.redis.com/staging/jira-doc-1540/modules/redisjson/redisjson-quickstart/)

### DIFF
--- a/content/modules/redisjson/redisjson-quickstart.md
+++ b/content/modules/redisjson/redisjson-quickstart.md
@@ -36,42 +36,69 @@ The `--raw` option forces the command to return raw output.
 
 ### Create JSON documents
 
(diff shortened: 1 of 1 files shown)


kaitlynmichael (dismissed):
I think this might benefit from numbered steps. Each action is hard to scan and separate from the others without any visual markers like step numbers or level 3 headers.

cmilesb (approved):

PR #2079, opened 2022-08-08
Update upgrading.md

yorozen:
1. rladmin status all - is a wrong command: 
rladmin status all
ERROR: invalid token 'all'

Should be:  `rladmin status extra all`

need to add what is the indicator for out of date database - OLD REDIS VERSION

### DIFF
--- a/content/rs/installing-upgrading/upgrading.md
+++ b/content/rs/installing-upgrading/upgrading.md
@@ -131,7 +131,8 @@ Before upgrading a database:
 
     - Use the admin console to open the Configuration tab for the database.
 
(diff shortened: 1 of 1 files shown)


yorozen:
Please see a customer complaint:

That's odd.  I didn't see any mention of a database upgrade required when going from 6.2.4 to 6.2.10 in the docs ([Upgrade an existing Redis Enterprise Software deployment | Redis Documentation Center](https://docs.redis.com/latest/rs/installing-upgrading/upgrading/)). 
Also, shouldn't there be an indicator if the database should be upgraded as per the docs?...
"Use the rladmin status all command to display configuration details. (An indicator appears in the command output when the database compatibility version is out-of-date.)"

rrelledge (approved):

PR #2104, opened 2022-08-18
RC: Update CIDR allow list to new UI

rrelledge:
Added new instructions, screenshots, and buttons and removed duplicated content.

[Staged preview](https://docs.redis.com/staging/release-rc-cidr/rc/security/cidr-whitelist/)


joniredis:
Hi @rrelledge!

I think there is a mixup between two separate features:

1) CIDR allow list at the database configuration level (that sets iptables firewall rules at the cluster level)
2) CIDR allow list for external accounts only (which I guess "If you use your own infrastructure to deploy Redis Cloud" refers to) at the subscription level under the "Connectivity" tab. This feature sets AWS Security Group rules at the VPC level. See screenshot below.

image


Unless I am missing something, it seems like you combined these two into this article at the database level? I think this page is still very useful, but it should be clearly described as "CIDR allow list" on the database configuration, and then I guess we should have a separate (perhaps linked) doc page for the "CIDR allow list" on the subscription level for external accounts.

WDYT?

rrelledge:
@joniredis Oh, the second allow list wasn't visible to me at the time. I will have to take another look at this and update this PR. I might need to send you some more details and questions to get this sorted out.

joniredis (approved):
@rrelledge looks good to me now!

PR #2121, opened 2022-08-26
JSON: Key name rules only apply to legacy path syntax

rrelledge:
[Staged preview](https://docs.redis.com/staging/jira-doc-1601/modules/redisjson/)


cmilesb (approved):
Looks good to me!

PR #2140, opened 2022-09-08
Avoid label selector in operator deployment verification

zcahana:
The existing documentation includes a command to verify the operator deployment is running, which uses a label selector for a label that isn't set on the operator deployment. This PR fixes the command to address the deployment by its name.


laurentdroin (approved):
LGTM

kaitlynmichael (approved):

PR #2150, opened 2022-09-18
Custom install platform support

adisht:
Hi, I've edited to reflect that custom install is supported for both RHEL 7 & RHEL 8 (and not just RHEL 7). I'd like Alon Shmueli to confirm before we change.


lanceleonard:
#hold-off until confirmation.

lanceleonard:
Confirmed.  Merging.

lanceleonard (approved):

PR #2154, opened 2022-09-20
Update centos-rhel-7-firewall.md

yorozen:
Remove the 7 since it's required to any RedHat


lanceleonard:
@yorozen - This PR duplicates an existing DOC ticket (DOC-1624) and should have been filed as a DOC ticket, as previously indicated.

Closing in favor of the preferred workflow.

PR #2170, opened 2022-09-28
DOC-1628: Wrong documentation for DB backup & export on RS

cmilesb:
Changed the AWS S3, GCP Storage, and Azure Blob sections of the backup and export database articles to better reflect the procedure for Software. Both pages are very similar - so you can make any suggestions for one and I'll carry them over onto the other one.

Staging links: 
https://docs.redis.com/staging/jira-doc-1628/rs/databases/import-export/export-data/#aws-simple-storage-service
https://docs.redis.com/staging/jira-doc-1628/rs/databases/import-export/schedule-backups/#aws-simple-storage-service

### DIFF
--- a/content/rs/databases/import-export/export-data.md
+++ b/content/rs/databases/import-export/export-data.md
@@ -147,99 +147,80 @@ To export to a local mount point:
 
 1. Verify that the user running Redis Enterprise Software has permissions to access and update files in the mount location.
 
-### AWS Simple Storage Service 
+### AWS Simple Storage Service
 
-To export data to an Amazon Web Services (AWS) Simple Storage Service (S3) [bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/creating-buckets-s3.html):
+To export data to an [Amazon Web Services](https://aws.amazon.com/) (AWS) Simple Storage Service (S3) bucket:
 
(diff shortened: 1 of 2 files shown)


antonum:
Please remove the mention of the root user key.  `For an AWS root user, see [Creating access keys for the root user](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_root-user.html#id_root-user_manage_add-key).`

Using and sharing the root user key severely violates security best practices. Instead, customers need to create a user/key with permissions limited to the specific bucket only.

cmilesb:
> Using and sharing the root user key severely violates security best practices. Instead, customers need to create a user/key with permissions limited to the specific bucket only.

Thanks @antonum - I'll take that out. 

Do you know what IAM user permissions need to be set for export and backup to work? I can link to [Creating IAM users](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html#id_users_create_console), but if there's a specific permission that needs to be set for the user, it would be nice to know that.

antonum:
You need to assign a policy, specific to the bucket to the IAM user you create. An example can be found here: https://docs.snowflake.com/en/user-guide/data-load-s3-config-aws-iam-role.html

kaitlynmichael (approved):

PR #2207, opened 2022-10-24
Updated OSS command & config compatibility

rrelledge:
Staged preview:
- [Command compatibility pages](https://docs.redis.com/staging/jira-doc-1732/rs/references/compatibility/commands/)
- [Config compatibility](https://docs.redis.com/staging/jira-doc-1732/rs/references/compatibility/config-settings/)


oranagra:
the diff is non-reviewable (i guess that's auto generated content, not actual source).
from a **quick** look at the staged content i see:
* the `busy-reply-threshold` config is missing, and p.s. considering it's an alias for `lua-time-limit` u guess it should have the same notes.
* the docs say that DUMP, RESTORE and OBJECT are unsupported, and that's wrong!
  maybe DUMP is unsupported in active active (i'm not sure, @AlonMagrafta?), but they're certainly supported in normal dbs.

please make the effort to run some diff between what's documented and the "facts" i posted in the ticket from the source code.
that's one of the reasons i was hesitated to point to specific examples of inconsistencies, since i didn't want it to look like i did the diff, i only noticed some issues, but i didn't bother to actually compare the lists.

one more thing:
i now realize that some configs in the source file i looked at are outdated (no longer supported by redis, since version 3.0 i think).
they should be removed from both the proxy and the docs.
these are all the `list-max*` configs, with the exception of `list-max-listpack-size` and `list-max-ziplist-size` which are the only two that are used. i.e. the ones ending with `entries` and `value` are to be removed.
i'll make sure to notify the proxy team to remove them too.

oranagra:
i asked the CRDT team. they do support DUMP.
but RESTORE is only supported for module keys, like ReJason (not for redis primitives)

rrelledge:
@oranagra Thanks for taking a look at this. I'll work on fixing the remaining inaccuracies and update this PR.

PR #2233, opened 2022-11-08
Modules command & config compatibility

rrelledge:
Staged previews:

- RediSearch
  - [commands](https://docs.redis.com/staging/release-module-compat/modules/redisearch/commands/)
  - [config](https://docs.redis.com/staging/release-module-compat/modules/redisearch/config/)

- RedisJSON
  - [commands](https://docs.redis.com/staging/release-module-compat/modules/redisjson/commands/)

- RedisGraph
  - [commands](https://docs.redis.com/staging/release-module-compat/modules/redisgraph/commands/)
  - [config](https://docs.redis.com/staging/release-module-compat/modules/redisgraph/config/)

- RedisTimeSeries
  - [commands](https://docs.redis.com/staging/release-module-compat/modules/redistimeseries/commands/)
  - [config](https://docs.redis.com/staging/release-module-compat/modules/redistimeseries/config/)

- RedisBloom
  - [commands](https://docs.redis.com/staging/release-module-compat/modules/redisbloom/commands/)
  - [config](https://docs.redis.com/staging/release-module-compat/modules/redisbloom/config/)


LiorKogan (commented):
I think that word 'Supported' in the 'Redis Enterprise' column can be removed. The green checkmark is enough.

Also when both "Flexible & Annual" and "Free and Fixed" are green checkmarks - consider removing both and just leaving a green checkmark with no text.

We would now have to sync these pages every time we make a documentation clarification, add a configuration parameter, etc. on redis.io . I'm concerned that we may miss syncs in the future...

lanceleonard:
> I think that word 'Supported' in the 'Redis Enterprise' column can be removed. The green checkmark is enough.
> 
> Also when both "Flexible & Annual" and "Free and Fixed" are green checkmarks - consider removing both and just leaving a green checkmark with no text.

For accessibility, I would prefer to leave the presentation as is.  That way, it's unambiguous for those a) using alternate experiences (screen readers) or b) those unable to render the images/colors.  Clarity is always preferable to ambiguity.

lanceleonard:
> We would now have to sync these pages every time we make a documentation clarification, add a configuration parameter, etc. on redis.io . I'm concerned that we may miss syncs in the future...

Yes, we have the same problem with the enterprise capability grid.  I believe automation would help here.  Until we're able to schedule that, it's a manual process.

PR #2235, opened 2022-11-09
Updated redisAccountMapping options

rrelledge:
[Staged preview](https://docs.redis.com/staging/jira-doc-1735/rc/security/single-sign-on/saml-sso/#add-saml-users)


joniredis:
@rrelledge I think the latest change is good. Just left wondering whether it would make sense to put the new XML part regarding the `redisAccountMapping` attribute in the section above it where we describe the assertion attributes? WDYT?

rrelledge:
@joniredis I made some changes to address your feedback. Can you take another look when you have some time?

joniredis:
@rrelledge yes that looks great! Please merge into the staging docs. One note: we'd probably have to update the screenshots for the SM config, since they changed slightly (the text itself already reflects these changes as we discussed them). The new UI is already available in the [QA environment](https://app-sm.k8s-idm.sm-qa.qa.redislabs.com/#/login

) that I think you used previously. If you would manage to update them before GA on the 20th that would be great, but no biggie if it follows some time soon after it reached production.

Thanks so much and have a great weekend!

PR #2252, opened 2022-11-23
Update routes.md

AhmedAlnoamany:
Adding the connection string via redis-cli

### DIFF
--- a/content/kubernetes/re-databases/routes.md
+++ b/content/kubernetes/re-databases/routes.md
@@ -52,3 +52,10 @@ Use the `openssl` command to access the database from outside the cluster.
   ```
 
   Replace the `<hostname>` value with the hostname for your new route.
+  
+  To connect via `redis-cli`:
+  
+  ```sh
+  redis-cli -h <hostname> -p 443 --tls --cacert ./ca.pem --sni <hostname>
+  ```
+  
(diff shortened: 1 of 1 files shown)


kaitlynmichael:
@AhmedAlnoamany thanks so much for this! Could you sign the Contributor License Agreement? I can push this as soon as I get an okay from R&D.

content/kubernetes/re-databases/routes.md
```diff
@@ -51,4 +51,11 @@ Use the `openssl` command to access the database from outside the cluster.
...
 
-  Replace the `<hostname>` value with the hostname for your new route.
\ No newline at end of file
+  Replace the `<hostname>` value with the hostname for your new route.
+  
+  To connect via `redis-cli`:
+  
+  ```sh
+  redis-cli -h <hostname> -p 443 --tls --cacert ./ca.pem --sni <hostname>
+  ```
```
zcahana, line 60: Maybe place this before the `openssl` example above? I would say the `redis-cli` example is more important and standard way of accessing a Redis DB, whereas the openssl is more of a sanity check used to test the connection.

zcahana (approved):
I've made one small suggestion, but otherwise this looks correct (also based on @laurentdroin's input).

PR #2295, opened 2023-01-05
Clarify deploying to existing VPC

joniredis:


joniredis:
@lanceleonard @rrelledge I wonder whether we should include deploying to an existing VPC as part of the main "create Flexible subscription" flow considering that this is only available to customers with cloud accounts.

joniredis:
Yes look good to me!

rrelledge (approved):

PR #2300, opened 2023-01-06
Update quick-start.md

bonghwankimRedis:
to help to get webhook file explicitly by using curl because no information about webhook file on the page.


kaitlynmichael:
@bonghwankimRedis I need to verify with the dev team if this is how we want customers to obtain the file, since it can't be used as is and must be modified. In the mean time, if you could sign the CLA agreement, that would make merging this easier. Thanks!

kaitlynmichael:
I've been informed that the webhook file is part of the bundle download, so customers don't need to do anything extra to obtain it if they already downloaded the bundle.

bonghwankimRedis:
thanks a lot. @kaitlynmichael

PR #2394, opened 2023-02-28
Update remove.md

maayanagranat:
Safe node removal - 6.4.2 March release content


kaitlynmichael (commented):
I left a few suggestions but I'm going to let @rrelledge do the approval since she's more familiar with this reference section.

rrelledge (approved):

PR #2418, opened 2023-03-08
DOC-1877: There is no need to merge multiple RDB files into a single RDB for import

cmilesb:
Staging link: https://docs.redis.com/staging/jira-doc-1877/rc/databases/import-data/


rrelledge (approved):
Looks good to me.

PR #2470, opened 2023-03-30
DOC-2064: Redis Cloud: missing instructions for active-passive when source is TLS enabled 

cmilesb:
Staging link: https://docs.redis.com/staging/jira-doc-2064/rc/databases/migrate-databases/#sync-using-active-passive


nermiller (commented):
@cmilesb a few comments unrelated to this specific PR, but rather the page as a whole:

* 2½ GB should be 2.5 GB
* Please add periods to complete sentences in the https://docs.redis.com/staging/jira-doc-2064/rc/databases/migrate-databases/#sync-using-active-passive
* If time, please try to reduce the number of button images. We can also say, "select Save", for example, and avoid using the word button. 
Otherwise, very nice work!

kaitlynmichael:
The PMs might know better than me, but I thought the name they wanted us to use for Active-Passive was Replica Of?

cmilesb:
> The PMs might know better than me, but I thought the name they wanted us to use for Active-Passive was Replica Of?

In software that's true, but in Cloud it is literally called "Active-Passive." I'll reach out to PM to see if there's any plan of changing that in the UI, but for now it's probably better to match the UI.

nermiller (approved):

PR #2526, opened 2023-04-27
Updates intro page, adds ingest and write-behind images

nermiller:


yaronp68:
@elena-kolevska please merge

PR #2570, opened 2023-05-19
RDSC-1260-fix-all-references-to-site-variables-in-rdi-latest

yaronp68:


elena-kolevska:
@yaronp68 should this go to master or rdi-latest?

PR #2581, opened 2023-05-26
Removed Nagios plugin article for 7.2

cmilesb:
Staging link: https://docs.redis.com/staging/DOC-2191/rs


amiramm (commented):
@rrelledge if we removed support for this completely and it is no longer available in any of the supported releases 6.2 and 6.4, then okay. If it is just deprecated, we should note it is deprecated and when support will be dropped. I'm not sure at this moment what's the case.

kaitlynmichael (approved):

PR #2590, opened 2023-06-02
RDI multi-targets and data_types

redima:


elena-kolevska:
I love that we have example yaml files for data transformation!

PR #2592, opened 2023-06-05
RDSC-296 update `map` block documentation with the accurate expression type and add example for nested jmespath

spicy-sauce:
… type and add example for nested jmespath

### DIFF
--- a/content/rdi/reference/data-transformation-block-types/map.md
+++ b/content/rdi/reference/data-transformation-block-types/map.md
@@ -5,19 +5,19 @@ description: Map a record into a new output based on expressions
 weight: 10
 alwaysopen: false
 categories: ["redis-di"]
-aliases: 
+aliases:
 ---
 
 Map a record into a new output based on expressions
 
 **Properties**
 
-| Name                          | Type     | Description                                   | Required |
-| ----------------------------- | -------- | --------------------------------------------- | -------- |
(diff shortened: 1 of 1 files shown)


yaronp68 (approved):

content/rdi/reference/data-transformation-block-types/map.md
```diff
@@ -5,19 +5,19 @@ description: Map a record into a new output based on expressions
...
 
 **Properties**
 
-| Name                          | Type     | Description                                   | Required |
-| ----------------------------- | -------- | --------------------------------------------- | -------- |
-| [**expression**](#expression) | `object` | Expression<br/>                               | yes      |
-| **language**                  | `string` | Language<br/>Enum: `"jmespath"`, `"sql"`<br/> | yes      |
+| Name                          | Type               | Description                                   | Required        |
+| ----------------------------- | ------------------ | --------------------------------------------- | --------------- | --- |
+| [**expression**](#expression) | `object`, `string` |                                               | Expression<br/> | yes |
```
elena-kolevska, line 17: @spicy-sauce Looks like we have an extra column space here?

spicy-sauce: good catch, the table is corrupted, I'll fix, thank you.

PR #2595, opened 2023-06-07
Update configure-acl.md

maayanagranat:
adding a note to avoid stuck state machine
should be merged with a PR for feature limitation i'll add soon

### DIFF
--- a/content/rs/security/access-control/rbac/configure-acl.md
+++ b/content/rs/security/access-control/rbac/configure-acl.md
@@ -32,7 +32,7 @@ Redis Enterprise lets you:
 
 - Include keys or key patterns with the "~" prefix.
 
-- Allow access to [pub/sub channels](https://redis.io/docs/manual/pubsub/) with the "&" prefix (only supported for databases with Redis version 6.2 and later).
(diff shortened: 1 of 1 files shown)


content/rs/security/access-control/rbac/configure-acl.md
```diff
@@ -32,7 +32,7 @@ Redis Enterprise lets you:
 
 - Include keys or key patterns with the "~" prefix.
 
-- Allow access to [pub/sub channels](https://redis.io/docs/manual/pubsub/) with the "&" prefix (only supported for databases with Redis version 6.2 and later).
+- Allow access to [pub/sub channels](https://redis.io/docs/manual/pubsub/) with the "&" prefix (only supported for databases with Redis version 6.2 and later). When adding permissions to specific channel, add 'resetchannels' at the begining to flush default configuration.
```
rrelledge, line 35:
```suggestion
- Allow access to [pub/sub channels](https://redis.io/docs/manual/pubsub/) with the "&" prefix (only supported for databases with Redis version 6.2 and later). When defining permissions for specific channels, include `resetchannels` at the beginning to flush the default configuration.
```

rrelledge (approved):
Since we already fixed this in 7.2 latest as part of PR #2668, I switched the base branch to 6.4-build so that this fix will also be included in the 6.4 docs.

PR #2618, opened 2023-06-15
chore: remove the outdated section from Uptrace integration

vmihailenco:
I've update some links and reworked the last section regarding notifications, because Uptrace is now able to send notifications out-of-the-box and does not require AlertManager.


kaitlynmichael (commented):
Looks good to me

PR #2636, opened 2023-06-21
K8s: RED-104136 add OLM admission controller note to AA

kaitlynmichael:
Added note the REAADB instructions explaining the admission controller and webhook are enabled by default on clusters created via OLM. 

Staged preview: https://docs.redis.com/staging/RED-104136/kubernetes/active-active/create-reaadb/


alon-zada (approved):
Great!

PR #2651, opened 2023-06-30
K8s: add note about alpha features flag

kaitlynmichael:
Jira: DOC-2454
Preview: https://docs.redis.com/staging/DOC-2454/kubernetes/active-active/prepare-clusters/
 https://docs.redis.com/staging/DOC-2454/kubernetes/active-active/


cmilesb (approved):

alon-zada (changes requested):
from previous descussion Yuval and I rather recommend to use the GA version since it contains additional features for AA and also bug fixes.
from where did this requirement or request to add a note regarding the preview notes came from?
@bascar wdyt?
cc: @yuvallevy2

kaitlynmichael:
@alon-zada request came from @bascar. I can add that we recommend using 6.4.2-6 and it's not supported before 6.2.4-4

bascar:
@kaitlynmichael Let's add something like: Since the A-A controller is now released as GA, we recommend upgrading to the latest release and not using any preview version as the solution has matured and any bugs were fixed.

However, someone might be stuck at the particular version and we still need the proper instructions for pre-GA usecases.

kaitlynmichael:
closing this. changes are now in https://github.com/RedisLabs/redislabs-docs/pull/2685

PR #2677, opened 2023-07-13
K8s: change redis.com to example.com

kaitlynmichael:
Jira: DOC-2471

### DIFF
--- a/content/kubernetes/re-clusters/connect-prometheus-operator.md
+++ b/content/kubernetes/re-clusters/connect-prometheus-operator.md
@@ -66,6 +66,6 @@ For more info about configuring the `ServiceMonitor` resource, see the [`Service
   - [Running exporters](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/running-exporters.md)
   - [Related resources](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#related-resources)
   - [Troubleshooting ServiceMonitor changes](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/custom-metrics-elements.png)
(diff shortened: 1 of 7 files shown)


content/kubernetes/re-clusters/connect-prometheus-operator.md
bascar, line 1: This file doesn't need changing, actually should be docs.redis.com and changing would break the links.

content/kubernetes/security/manage-rec-credentials.md
bascar, line 1: This change is ok, but the data in line 35, the map, actually has bad data for password and since we are changing the output, we should fix in this example. Should be:
map[password:MTIzNDU2NzgK username:ZGVtb0BleGFtcGxlLmNvbQo=]

Also, line 41, with the bash,  the password decoded should now be(the example is actually wrong)
echo MTIzNDU2NzgK | base64 --decodexc

content/kubernetes/re-clusters/connect-prometheus-operator.md
```diff
@@ -66,6 +66,6 @@ For more info about configuring the `ServiceMonitor` resource, see the [`Service
   - [Running exporters](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/running-exporters.md)
   - [Related resources](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#related-resources)
   - [Troubleshooting ServiceMonitor changes](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/custom-metrics-elements.png)
-- docs.redis.com
+- docs.example.com
```
kaitlynmichael, line 69:
```suggestion
- docs.redis.com
```

PR #2698, opened 2023-07-24
DOC-2515 speedb changes

kaitlynmichael:
Jira: DOC-2515

### DIFF
--- a/content/kubernetes/re-clusters/auto-tiering.md
+++ b/content/kubernetes/re-clusters/auto-tiering.md
@@ -39,7 +39,7 @@ To deploy a Redis Enterprise cluster (REC) with flash storage, you'll need to sp
 
 - enable Auto Tiering (`enabled: true`)
 - flash storage driver (`flashStorageEngine`)
-  - `rocksdb` or `speedb`
+  - `rocksdb` or `speedb`(default)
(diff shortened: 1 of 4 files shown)


content/kubernetes/re-clusters/auto-tiering.md
```diff
@@ -73,7 +73,7 @@ Below is an example REDB custom resource:
 apiVersion: app.redislabs.com/v1alpha1
 kind: RedisEnterpriseDatabase
 metadata:
-  name: rof-redb
+  name: autoteiring-redb
```
AlonMagrafta, line 76: @bascar can you confirm this?

AlonMagrafta (commented):
LGTM, I left @bascar one question to confirm.
@kaitlynmichael Thanks. If you also know the answer then we can proceed :)

content/kubernetes/re-clusters/auto-tiering.md
bascar: @AlonMagrafta @kaitlynmichael API is not changing, so this should not change.

bascar: Actually, this is just a name, with no impact. Can stay this way, sorry.

PR #2754, opened 2023-08-15
RS: Troubleshoot DB connectivity guide

rrelledge:
Adapted from a troubleshooting guide provided by support. Requested in DOC-670.

[Staged preview](https://docs.redis.com/staging/DOC-670/rs/databases/connect/troubleshooting-guide/)


yoavredis (approved):
Looks OK to me.
Cluster PMs should review this too.

AlonMagrafta:
Nice page!
What I'm mostly missing is the following:
For some of the topics on the page, there are instructions to verify something (e.g. parameter under 80%) but it doesn't say anything about what to do if it's above 80%. Even if it seems obvious, I think we should write it. Also if it's "contact support" at that phase. 
I do think we can add later on, not a blocker for this PR

AlonMagrafta:
@gilbennun ^^

PR #2782, opened 2023-08-22
remove old gears links

yaronp68:


yaronp68:
needs to be backported to latest

spicy-sauce (changes requested):
```> Note: If your Redis Enterprise Cluster is version 6.2.18 or higher, please download the RedisGears bundle that includes the Python plugin.
```
Now that you have only one link per distribution, this seems to be irrelevant

spicy-sauce (approved):

PR #2794, opened 2023-08-24
K8s: modules limitation - add warning to recover and upgrade pages

kaitlynmichael:


bascar (approved):
LGTM!!

laurentdroin:
It's not just about manually uploaded modules. It's if you have any database using a module that doesn't come bundled with the RS version currently being used. It might happen with modules that were not manually uploaded. For example, if a recovery is needed right after an upgrade.

PR #2809, opened 2023-08-30
Update rs-7-2-4-52.md

michaelatredis:
I've confirmed both working with a customer and by checking the CCS of an actual RHEL-7 environment. The correct os_name field should be "rhel" not "rhel7". If it is set to "rhel7", modules will fail to load citing "Error:  combination of OS rhel77 [. . .] is not supported".


rrelledge:
This fix was also included as part of #2813, so I am going to close this PR.

PR #2842, opened 2023-09-14
RDSC-1499: support for redis lookup transformation block

vd2org:


spicy-sauce:
@vd2org you should rebase it as it was forked from the `master` and should be merged to `rdi-preview-build`

spicy-sauce (approved):

redima:
so do we expect a data engineer to be a redis expert now? I remember this was a problem and the similar syntax proposal got rejected when we worked on `redis.write` block.

redima:
Also, I think it should be limited to a set of specific read commands, as currently I can delete of modify data by passing `DEL` as a command.

spicy-sauce (approved):

spicy-sauce:
> so do we expect a data engineer to be a redis expert now? I remember this was a problem and the similar syntax proposal got rejected when we worked on `redis.write` block.

We can just wrap it with the supported commands, for example, `hash` instead of `hgetall` and `hget` with the field and key.
Not sure if this should be handled at this point, but probably such an abstraction is preferred. 
@yaronp68 WDYT?

spicy-sauce:
> Also, I think it should be limited to a set of specific read commands, as currently I can delete of modify data by passing `DEL` as a command.

I agree, see my answer above, if we support only `hash`, `json` and other read commands, this won't be available.

yaronp68:
> > so do we expect a data engineer to be a redis expert now? I remember this was a problem and the similar syntax proposal got rejected when we worked on `redis.write` block.
> 
> We can just wrap it with the supported commands, for example, `hash` instead of `hgetall` and `hget` with the field and key. Not sure if this should be handled at this point, but probably such an abstraction is preferred. @yaronp68 WDYT?

I think we can just specify the expected key type: Hash, JSON, Set etc and not the command. Obviously only a handful of read commands (and some members commands)

yaronp68:
Full list of commands Lookup should support and the value user should specify:
HGETALL - Hash
JSON.GET - JSON
SMEMBERS - Set.All
ZRANGE - SortedSet.Range AND ZRANGE with BYSCORE - SortedSet.Score
ZPOPMAX - SortedSet.Max
ZPOPMIN - SortedSet.Min
GET - String
MGET - Strings
(we can add a substring command later)

spicy-sauce:
> Full list of commands Lookup should support and the value user should specify: HGETALL - Hash JSON.GET - JSON SMEMBERS - Set.All ZRANGE - SortedSet.Range AND ZRANGE with BYSCORE - SortedSet.Score ZPOPMAX - SortedSet.Max ZPOPMIN - SortedSet.Min GET - String MGET - Strings (we can add a substring command later)

Let's have a new issue for this, as it was already implemented like this (this is the adjustments in the documentation).

content/rdi/reference/data-transformation-block-types/lookup.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+Get data from another key
+
+**Properties**
+
+| Name                             | Type           | Description               | Required |
+| -------------------------------- | -------------- | ------------------------- | -------- |
+| [**connection**](#option1fields) | `string`       | Connection name<br/>      | yes      |
+| [**cmd**](#option1fields)        | `string`       | Command to execute<br/>   | yes      |
```
yaronp68, line 18:
```suggestion
| [**cmd**](#option1fields)        | `string`       | Redis command to execute<br/>   | yes      |
```

content/rdi/reference/data-transformation-block-types/lookup.md
```diff
@@ -0,0 +1,41 @@
...
+| [**args**](#option1fields)       | `list[string]` | Command's parameters<br/> | yes      |
+| [**language**](#option1fields)   | `string`       | Expression language<br/>  | yes      |
+| [**field**](#option1fields)      | `string`       | Target field<br/>         | yes      |
+
+**Additional Properties:** not allowed
+
```
yaronp68, line 24: Add list of allowed commands (even if we are not validating these can only be commands that do read of one or more keys / members
Also we should document how different commands response will look like in a field

yaronp68 (changes requested):

yaronp68 (approved):

PR #2918, opened 2023-11-06
K8s: Sharding disabled with 1 shard REAADB

kaitlynmichael:
Jira: DOC-3054
Staging: https://docs.redis.com/staging/DOC-3054/kubernetes/active-active/create-reaadb/


alon-zada (approved):
Great!

PR #2932, opened 2023-11-13
DOC-2923 Feedback from Docs Bash.

amberdoctor-redis:
https://redislabs.atlassian.net/browse/DOC-2923


rrelledge (approved):
Looks good to me.

PR #2939, opened 2023-11-13
DOC-2823 Fixed phrasing to improve clarity per docs bash feedback.

amberdoctor-redis:
https://redislabs.atlassian.net/browse/DOC-2823

### DIFF
--- a/content/rs/installing-upgrading/install/plan-deployment/persistent-ephemeral-storage.md
+++ b/content/rs/installing-upgrading/install/plan-deployment/persistent-ephemeral-storage.md
@@ -40,7 +40,7 @@ for the persistent volume.
 
 ## Ephemeral storage
 
-Ephemeral storage is optional. If configured, the cluster stores in the ephemeral storage any temporary information that does not need to be persisted.
+Ephemeral storage is optional. If configured, temporary information that does not need to be persisted is stored by the cluster in the ephemeral storage.
 This improves performance and helps reduce the load on the persistent storage.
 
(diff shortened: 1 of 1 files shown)


content/rs/installing-upgrading/install/plan-deployment/persistent-ephemeral-storage.md
```diff
@@ -40,8 +40,7 @@ for the persistent volume.
 
 ## Ephemeral storage
 
-Ephemeral storage is optional. If configured, the cluster stores temporary information
-that does not need to persist in the ephemeral storage.
+Ephemeral storage is optional. If configured, the cluster stores in the ephemeral storage any temporary information that does not need to be persisted.
```
rrelledge, line 43: Word order suggestion
```suggestion
Ephemeral storage is optional. If configured, the cluster stores temporary information that does not need to be persisted in the ephemeral storage.
```

amberdoctor-redis: The suggestion puts the phrasing back to what it was and changes the meaning. 

In the original version, "in the ephemeral storage" might be mistaken for modifying persisted.  

We could try this?

Ephemeral storage is optional. If configured, the cluster stores - in the ephemeral storage - temporary information that does not need to be persisted.

Or (but it changes it to a more passive tense - way clearer though)

Ephemeral storage is optional. If configured, temporary information that does not need to be persisted is stored by the cluster in the ephemeral storage.

PR #2994, opened 2023-11-30
Updates to the networking RS section

ggeorges2015:
Hello! Please note that I created a new Route 53 page and wanted your feedback on it before overwriting the current one. The current one is too verbose and based on old screenshots which I changed. 

We would have to take that content if approved and overwrite the current one.

includes: DOC-3017 DOC-3026


ggeorges2015 (commented):
Accepted Kaitlyn's suggestions

kaitlynmichael (approved):

PR #2999, opened 2023-12-03
Update upgrade-database.md

adisht:
Users perceive DB upgrade as an easy straightforward task. For now I suggest to remove the "upgrade policy" explanations and leave the redirection to previous releases. There's still work to do on this page to simplify the process, especially with the backup explanations.


rrelledge (approved):
Looks good to me.

PR #3059, opened 2024-01-03
Add tags to all instances referencing the Debezium

galilev:
Server image.
pulling the image from quay instead of Docker

### DIFF
--- a/content/rdi/installation/pacemaker.md
+++ b/content/rdi/installation/pacemaker.md
@@ -5,7 +5,7 @@ description: Deploy a Debezium Server VM cluster using Pacemaker
 weight: 74
 alwaysopen: false
 categories: ["redis-di"]
-aliases: 
+aliases:
 ---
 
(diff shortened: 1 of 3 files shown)


content/rdi/installation/pacemaker.md
```diff
@@ -343,7 +344,7 @@ Verify the container is running with `sudo podman ps -a`. This should result in
 
 ```shell
 CONTAINER ID   IMAGE             COMMAND              CREATED          STATUS          PORTS     NAMES
-2aecd4e9fa31   debezium/server   "/debezium/run.sh"   2 minutes ago    Up 2 minutes              dbz_server
+2aecd4e9fa31   quay.io/debezium/server:{{<param rdi_debezium_server_version>}}   "/debezium/run.sh"   2 minutes ago    Up 2 minutes              dbz_server
```
spicy-sauce, line 347: This seems not indented correctly (even when replacing the `{{<param rdi_debezium_server_version>}}` with actual version. Try to replace it with 2.4.0 for example and see that it's indented as before, then put the placeholder.

spicy-sauce (changes requested):

spicy-sauce (approved):

PR #3076, opened 2024-01-11
K8s: change delete pods argument to avoid force delete

kaitlynmichael:
change --force to --now to avoid force deleting pods. 
DOC-3198

### DIFF
--- a/content/kubernetes/re-clusters/cluster-recovery.md
+++ b/content/kubernetes/re-clusters/cluster-recovery.md
@@ -44,18 +44,6 @@ The Redis Enterprise for Kubernetes automates these recovery steps:
     kubectl patch rec <cluster-name> --type merge --patch '{"spec":{"clusterRecovery":true}}'
     ```
 
-    {{< note >}}
-In some cases, pods do not terminate when the statefulSet is scaled down as part of the cluster recovery.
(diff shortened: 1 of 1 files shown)


content/kubernetes/re-clusters/cluster-recovery.md
```diff
@@ -51,7 +51,7 @@ If pods are stuck in `terminating` or `crashLoopBack` and do not terminate grace
 To work around this, delete the pods manually with:
 
 ```sh
-kubectl delete pods <pod> --grace-period=0 --force
+kubectl delete pods <pod> --now
```
zcahana, line 54: Despite what I wrote in the ticket, I've realized in recent weeks that re-deleting with `--now` doesn't help either in case pod termination is already stuck. This is, in fact, a longstanding bug in Kubernetes, but without a fix in place we shouldn't be recommending users to do it.

Also, apparently the operator itself is force deleting RS pods as part of the recovery process. We'll be adding some operator logic to eliminate or reduce the bad consequences of force deletion.

The upside of that, is that since the operator is force deleting the RS pods, there isn't really a scenario in which _"pods do not terminate when the statefulSet is scaled down as part of the cluster recovery"_, and hence this entire `{{< note >}}` can be removed (I'm guessing it was written before we had the operator force delete as part of recovery).

PR #3100, opened 2024-01-29
RS 7.4 version selector updates - 6.4-build branch

rrelledge:
DOC-3076

Staged preview: https://docs.redis.com/staging/DOC-3076-6.4/rs/

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -78,7 +78,8 @@
                     <span class="menu__version-selector__toggler closer version-selector-control">&#x25B2;</span>
                 </button>
                 <div class="menu__version-selector__list version-selector-control">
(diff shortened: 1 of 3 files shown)


mich-elle-luna (approved):
I think it is fine based on the raw file, but it does wrap in the preview where (latest) appears on a second line).

rrelledge:
> I think it is fine based on the raw file, but it does wrap in the preview where (latest) appears on a second line).

@mich-elle-luna I think this is because the version selector box automatically seems to resize to fit the smaller text on the branches that do not have `latest` in the name. I'm not sure of a quick and easy way to fix it right now. Attempting to set a minimum width didn't seem to change anything.

PR #3153, opened 2024-02-27
RS: Clarify RS v5.4.2 & later use CRDB protocol version 1

rrelledge:
DOC-3356

[Staged preview](https://docs.redis.com/staging/DOC-3356/rs/installing-upgrading/upgrading/upgrade-active-active/)


mich-elle-luna (approved):
add However on line 19, otherwise this looks good, thank you!

PR #3180, opened 2024-03-08
Wrong tag for rhel7 openshift images

mzervakis:


kaitlynmichael (approved):
Thanks for catching that!

PR #3188, opened 2024-03-12
Update memtier-benchmark.md

mar1boroman:
Changed a possible typo from 'Four masters and 2 replicas' to '2 masters and 2 replicas'


rrelledge (approved):
Thank you!

